PR #15, opened 2016-09-17
README.md: add Travis CI shields.io badge

codygarver:


danrabbit:
I don't see the point of this. We require the build to pass in order to merge, so master will never fail to build

PR #45, opened 2016-12-18
fix flat button focus style in a selected row

danrabbit:
Edge cases, man. Edge cases


snwh:
Or lack of edge cases. ;)

PR #46, opened 2017-01-04
Default decoration

danrabbit:
This branch changes the style of `default-decoration` to match that of `titlebar.compact`.

Up until this point, the stylesheet has been mostly built around the assumption that a titlebar has a toolbar below it. However, this isn't always the case. Changing the style in this way should make some non-native apps look nicer as well as not require the explicit use of `Gtk.Headerbar` if no toolbar is needed.

Please test this branch thoroughly for adverse effects before merging :)

**Before:**

screenshot from 2017-01-04 15 17 57

**After:**

screenshot from 2017-01-04 15 17 41

### DIFF
--- a/gtk-3.0/apps.css
+++ b/gtk-3.0/apps.css
@@ -289,6 +289,12 @@ ConversationListView.view.cell:selected:focus {
     background-color: alpha(#000, 0.5);
 }
 
+.panel menubar,
+.panel .menubar {
+    box-shadow: none;
+    border: none;
+}
+
 .composited-indicator > revealer label,
 .composited-indicator > revealer image,
 .composited-indicator > GtkRevealer {
--- a/gtk-3.0/gtk-widgets.css
+++ b/gtk-3.0/gtk-widgets.css
@@ -2316,7 +2316,7 @@ radio:checked:focus,
 toolbar,
 .toolbar {
     -GtkWidget-window-dragging: true;
-    padding: 0 3px 3px;
+    padding: 3px;
     border: 1px solid shade (@titlebar_color, 0.56);
     border-width: 0 0 1px;
     box-shadow: inset 0 -1px 0 0 alpha (@bg_highlight_color, 0.2);
--- a/gtk-3.0/gtk-widgets.css
+++ b/gtk-3.0/gtk-widgets.css
@@ -2111,7 +2111,7 @@ menubar,
 menubar:backdrop,
 .menubar:backdrop {
     background-color: shade (@titlebar_color, 0.98);
-    border: 1px solid shade (@titlebar_color, 0.68);
+    border-color: shade (@titlebar_color, 0.78);
 }
 
 .titlebar .menubar {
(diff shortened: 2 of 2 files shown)


Defman21:
Firefox looks ugly with this change :( I assume this line is for Alt-key menus of different applications.

Before:
image

After:
image

Defman21:
Also unfocused windows looks weird. Is there a way you can fix that? :)

image

danrabbit:
non-native apps are just never going to look like native apps. This change prioritizes native gtk3 apps over gtk2 and non-native apps

Defman21:
I've not asked to make non-native apps looks like native, I doubt it's possible. I just want to know if the issue with tabs in the unfocused Firefox is an issue of the theme or Firefox itself.

cassidyjames:
@Defman21 Firefox is not a true GTK app, so it is not picking up the `:backdrop` styles like it should. That is a bug in Firefox and not something our stylesheet can address.

harisvsulaiman:
@cassidyjames I know design is subjective, chrome firefox and every non-native App looks odd to me.

ricvelozo:
@Defman21 I found an addon to hide the title bar: https://addons.mozilla.org/pt-br/firefox/addon/hide-caption-titlebar-plus-sma/

P.s.: The default “Firefox” orange button can be removed.

Gabriel-p:
Could this change be the reason I now see a top bar on some apps which I could *swear* I didn't see before? This is a "Save window" from Sublime Text 3 for example:

captura realizada el 2017-01-24 11 57 49

Was that top bar always there?? First, it makes no sense having a "Close button" on such a window. Second, I am almost certain that wasn't there a few days ago. Am I going mad?

cassidyjames:
@Gabriel-p as far as I know, Sublime (and other apps using legacy toolkits such as GTK2) has always had a titlebar on its dialog. The titlebar nor close button were added; however, the styling of the titlebar on legacy apps has been updated as a result of this change.

We are prioritizing the look and consistency of newer, more modern apps written with GTK3 over keeping older, legacy apps written in GTK2 looking the same as they always have.

Gabriel-p:
Then it must be me then. I could I sworn that that titlebar (also shows in Okular) was not there before. Thanks @cassidyjames.

Gabriel-p:
Wait, no. I am 200% sure that this titlebar was not there in Okular before.

captura realizada el 2017-01-25 12 14 16

@cassidyjames are you saying that was there since the initial release of Loki?

cassidyjames:
The titlebar itself was there, it just looked flatter. :wink:

Gabriel-p:
It's settled then, I'm going mad. Thanks @cassidyjames.

rootsandculture:
screenshot from 2017-01-27 14_49_37
screenshot from 2017-01-30 18 46 15
Does it happen for this branch? Latest update makes ie. Firefox looks non-integrated. The another snapshot is before that update.

PR #70, opened 2017-01-28
Give stackswitcher image buttons in titlebars a border

### LINKED ISSUE #69: Gtk.StackSwitcher in a headerbar is missing border
Hi there, in Gtk applications that use StackSwitcher with icons, the apearance is quite odd, as you can see in the attachment, it misses the borders and the selected icon has a ellipsis around it. Perhaps it should look like Granite ModeButton? I don't know if this as to do with a proper Gtk style class or something...

Using Gtk.StackSwitcher:
stack

Granite.ModeButton:
modal

Tested in **Elementary OS Loki 0.4 with GTK 3.18**

danrabbit:
Fixes #69 

Just some ridiculous specificity stuff we have to do because `:not` isn't a thing in gtk3.18

### DIFF
--- a/gtk-3.0/gtk-widgets.css
+++ b/gtk-3.0/gtk-widgets.css
@@ -1053,7 +1053,8 @@ GtkComboBox.combobox-entry .button {
 button,
 actionbar button.text-button,
 .button,
-.action-bar .button.text-button {
+.action-bar .button.text-button,
+.titlebar .stack-switcher .button.image-button {
     text-shadow: 0 1px @text_shadow_color;
     icon-shadow: 0 1px @text_shadow_color;
     background-image:
@@ -1123,14 +1124,16 @@ toolbar button,
 }
 
 button:hover,
-.button:hover {
+.button:hover,
+.titlebar .stack-switcher .button.image-button:hover {
     color: @text_color;
 }
 
 button:focus,
 popover actionbar button:focus,
 .button:focus,
-.popover .action-bar .button:focus {
+.popover .action-bar .button:focus,
+.titlebar .stack-switcher .button.image-button:focus {
     color: @text_color;
     border-color: @colorAccent;
     box-shadow:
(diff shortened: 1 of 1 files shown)


ajcerejeira:
@danrabbit Thanks for the quick reply and fix, I copied that stylesheet you referenced in the pull request to my elementary theme folder and i tested again, and it seems to work!

When I open a GTK app i get the following warning (probably my mistake, perhaps i should have cloned this repo, instead of changing only the `gtk-widgets.css`)
```
(demo:2685): Gtk-WARNING **: Theme parsing error: gtk-widgets.css:2046:14: 'min-height' is not a valid property name

(demo:2685): Gtk-WARNING **: Theme parsing error: gtk-widgets.css:2047:13: 'min-width' is not a valid property name

(demo:2685): Gtk-WARNING **: Theme parsing error: gtk-widgets.css:2052:13: 'min-width' is not a valid property name
```
Just a very small detail: when the headerbar is taller (for example when a button uses the `LARGE_TOOLBAR` size) the `StackSwitcher` also grows, and I think this is not the desirable behavior:

captura de ecra de 2017-01-28 22 55 24

captura de ecra de 2017-01-28 22 53 32

danrabbit:
Nope, the min-height and min-width part is in master and it's just a warning we have to deal with right now to be compatible with both gtk 3.18 and 3.20. Nothing you did wrong :)

The widget height is actually something that you control in-app. Gtk defaults to `valign = Gtk.Align.FILL`. So if you don't want it to grow, you need to set `valign = Gtk.Align.CENTER` in your app.

ajcerejeira:
@danrabbit Thanks a lot, you were right, everything works now as expected!

> Just some ridiculous specificity stuff

I know this is a very specific use case, but i think this way the consistency between Granite and Gtk apps gets better :)

danrabbit:
@ajcerejeira Glad it's working as you expect :) Hopefully we can clean up some of these specificity issues with the `:not` selector in the future

PR #77, opened 2017-03-20
Fixed switch active\inactive state background, fixed selection in entry and textview

sourenaraya:
fixes #62 and #59

PR #85, opened 2017-03-25
Revert change that breaks switch active states in 3.18

danrabbit:
Revert changes made in #77 to the selector that targets 3.18


sourenaraya:
if backward compotability is needed, maybe we just keep :checked and :active together?

danrabbit:
No worries! I reviewed it so it's my bad 😉

PR #90, opened 2017-04-04
Pulse Animation: Pulse 5 times, then hold color

Philip-Scott:
GTK freaks out and cpu goes to 100% if it tries to animate something when the screen is off. This stops the animation after 5 pulses, and holds the color


danrabbit:
I think I'm -1 on this. It should be the job of your application to remove the animation when you don't want it anymore

PR #99, opened 2017-06-07
keep maximized windows round and draw black behind them

danrabbit:
Solves two problems:
1. non-native apps don't seem to pick up the maximized class so they're always rounded. Now everything is always rounded
2. wallpaper color won't poke through because we're smart and draw a black box behind the window when it's maximized

screenshot from 2017-06-07 11 17 51


snwh:
That'll do

PR #112, opened 2017-07-07
Fix separators

danrabbit:
Fixes menu separators by changing padding to margin
removes deprecated values
draw seps with magic gradients to workaround gtk align fill


tintou:
There used to be 0 gap between the separator and the other widgets, here you can see that the separator has a margin (maybe related to the min-width of 4px)
Issue

PR #113, opened 2017-07-08
Clean up label selectors

danrabbit:
Make sure we're selecting with `label`


gtk-3.22/gtk-widgets.css
```diff
@@ -3557,7 +3528,7 @@ GtkDialog .source-list {
 }
 
 .sidebar .label,
-.sidebar-item GtkLabel {
+.sidebat-item label {
```
eyelash: typo?

danrabbit: Oops, nice catch!

PR #124, opened 2017-07-09
Make pane separator hitboxes big again

### LINKED ISSUE #56: [$15] Extra Margin in GtkPaned
Currently we get a Border around GtkPaned windows (8px right and bottom)

Relevant screenshot:
bildschirmfoto von 2017-01-23 20 59 52

This can be fixed if we remove that margin in gtk-widgets.css. But i have no idea what was the reason in the first place we defined it like that

danrabbit:
Fixes #56


tintou:
It's better but the separator isn't centered on the hitbox

danrabbit:
This is intentional. If you center it, it will cover scrollbars

tintou:
Mm, you're right for the scrollbar, but can't we have like 1px on the left so that it looks more natural ?

danrabbit:
This is how it was before. I'm concerned about making a change here since we did that deliberately in the past

PR #127, opened 2017-07-10
Clean up notebooks

danrabbit:
* Remove old selectors from 3.18
* Replace tab overlap with negative margin
* Fix border for unchecked tabs
* Fix backdrop for checked tabs


tintou:
There seems to be a tiny regression here, if you look at gtk3-widget-factory and focus the second tab, you'll see that the third tab is moved from 1px to the right, it wasn't the case before.

PR #153, opened 2017-07-22
Rewrite menuitems and modelbuttons

danrabbit:
* Remove old 3.18 `GtkModelButton` selector
* Apply the same styles for `menuitem`, `.menuitem`, and `modelbutton` no matter their container (`popover`, `popup`, or `menu`)
* Selectors in alphabetical order
* Fix accelerator selector

### DIFF
--- a/gtk-3.22/gtk-widgets.css
+++ b/gtk-3.22/gtk-widgets.css
@@ -1049,7 +1049,7 @@ button:hover,
     color: @text_color;
 }
 
-button:focus,
+button:not(.menuitem):focus,
 popover actionbar button:focus,
 .button:focus,
 .popover .action-bar .button:focus,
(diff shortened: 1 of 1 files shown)


gtk-3.22/gtk-widgets.css
```diff
@@ -1909,88 +1909,54 @@ menubar:backdrop,
...
-}
-
-.menuitem .accelerator {
-    color: alpha (@text_color, 0.4);
-}
-
-.menuitem .accelerator:hover,
-.menuitem .accelerator:active {
-    color: alpha (@fg_color, 0.3);
+modelbutton {
```
davidmhewitt: Is it right to select just the modelbutton here? Considering we're still in the menu part of the stylesheet. The properties set on it do not seem to be consistent with other button selector properties in the rest of the stylesheet.

PR #171, opened 2017-08-10
Add a style for .keycap

danrabbit:
screenshot from 2017-08-10 12 08 42


cassidyjames:
Something that we realized when doing this [styling for the System76 website](http://support.system76.com/articles/terminal-basics/) is that most keyboards have black keycaps these days; would it be worth doing that, or you good with the white?

danrabbit:
The black seems kind of aggressively contrasty to me

cassidyjames:
:woman_shrugging: fair

Issue #189, opened 2017-08-26
Bring brand/palette colors in as constants


cassidyjames:
...so we can use them instead of hardcoded values.

danrabbit:
Closing since this was fixed for 3.22

PR #197, opened 2017-09-05
Replace colors in gtk.css with palette colors

danrabbit:
First commit towards fixing #189 

* Replace values in gtk.CSS with colors from the palette
* Remove dead color constants


snwh:
I think maybe `#FFF` should be defined as white or something too

danrabbit:
Gonna merge because Juno cycle #YOLO

PR #202, opened 2017-09-14
fix text color and shadow in titlebar back buttons

danrabbit:


lainsce:
I can confirm this fixes the back button for Palaura's use case
Imgur

danrabbit:
Gonna use my admin super powers to merge since Paulo confirms the fix and Sam is off to Uni these days ;p

PR #203, opened 2017-09-15
Fix padding for sidebar stack switcher

danrabbit:
Fixes the height of a sidebar stack switcher for gtk 3.18 (Loki)


danrabbit:
I'm gonna close this since it doesn't seem like we're going to release the stylesheet or Code on Loki again

PR #208, opened 2017-09-27
reduce the height of the panel

danrabbit:


cassidyjames:
Is this fixing a bug, or just a design change?

lainsce:
@cassidyjames In a "franken-juno" install, wingpanel is using a little more space than usual. This is fixing this UI error. Unless I'm wrong about this PR.

danrabbit:
Yeah that's it. Here's a screenshot of Loki. I'll double check that it's the same with this branch in just a second...

screenshot from 2017-09-27 14 59 19

danrabbit:
Gonna merge because Juno cycle #YOLO

PR #214, opened 2017-10-13
Add a style for the Terminal widget in the installer

danrabbit:


danrabbit:
Gonna add this as a fallback style in the installer instead

danrabbit:
Closing since the PR to the installer was merged

PR #236, opened 2017-11-29
Add a style for panel.color-light.translucent

danrabbit:
Doesn't do anything until wingpanel supports this


cassidyjames:
Here's a pretty high-contrast wallpaper with the panel at 85%:

testing-white-panel

PR #248, opened 2017-12-31
Added Silver100 to temperature scale

Philip-Scott:
Adds a Silver100 stop to the temperature scale to better reflect how temperature is being visualized on other platforms. Plus, the white just makes the transition look better :wink: 

Example: 
- http://content.hwigroup.net/images/products_xl/169930/18/philips-hue-connected-bulb-starter-pack-e27.jpg

## Before
screenshot from 2017-12-30 20 17 34

## After
screenshot from 2017-12-30 20 17 01


lewisgoddard:
This seems like a much more realistic expression of how this will actually effect color. +1

PR #257, opened 2018-01-16
misc dark style fixes

danrabbit:
* Use a color from the palette for text color
* Get rid of legacy selector `EggListBox`
* Use `mix` for toolbar  and titlebar image colors that work better on light and dark stylesheets
* Make the selector for toolbar and titlebar images less specific so that it keeps buttons and plain Gtk.Images matching


ztefn:
There's a typo on [line 2294 in gtk-widgets.css](https://github.com/elementary/stylesheet/pull/257/files#diff-2fb0e82ca0266ef5ad606424d133551bR2294) which reads '.titlbar' instead of '.titl**e**bar'.

danrabbit:
@ztefn fixed! Thanks :)

PR #260, opened 2018-01-19
Fix menuitem image placement

danrabbit:
Gtk.ImageMenuItem is now deprecated and the docs recommend packing images inside a container with the label. The problem is this fucks up the alignment, so here we do some CSS magic to fix it


jeremypw:
Does RTL not require any fix?

danrabbit:
It does but I’m having trouble with it for some reason. I figure fixing half the problem is better than not fixing the problem 😬

PR #269, opened 2018-02-10
Fix GtkScale rounded borders and margin, show fill level

markodp:
* round correct corners for RTL or inverted widgets
* remove highlight margin for inner border for more accurate Granite Seekbar
* display fill level property, when set

before:
screenshot from 2018-02-10 14 34 22

after:
screenshot from 2018-02-10 14 34 34

This only fixes 3.22 version. There are similar issues for gtk-3.0. Is that version still active and should it be fixed? I haven't checked 2.0.

I'm not sure if style for fill level is ok, i just added some color to it so that it's displayed. And also since highlight is partly transparent, fill level makes it more dark.

Let me know if anyone wants code for this test app.

### DIFF
--- a/gtk-3.22/gtk-widgets.css
+++ b/gtk-3.22/gtk-widgets.css
@@ -824,19 +824,39 @@ scale.vertical mark indicator {
 }
 
 scale highlight {
-    border-radius: 12px 0 0 12px;
     background-color: alpha (#000, 0.45);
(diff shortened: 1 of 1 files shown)


danrabbit:
I don't think Loki will get any more stylesheet updates, so I wouldn't worry about the styles in the `gtk-3.0` folder

PR #271, opened 2018-02-14
Fix linked border error/focus states

### LINKED ISSUE #76: Missing linked entry borders on focus/error
This is because of how the linked entries drop borders. We could look at faking it with an inset box shadow or something if possible.

focus-and-error-linked-entry

davidmhewitt:
Fixes #76 

image

It was a specificity issue that was preventing the borders from showing on some of the sides of the linked buttons.

So, this seems like a really nasty way to fix this since it essentially overrides all of the specificities I could think of with the error/focus border state. Maybe there's a better way of doing this, but I only dabble with CSS.

Trying to fake it with an inset shadow was no good since it did what an inset shadow says on the tin, you ended up with a second inset border.


cassidyjames:
…and this is one of those times I wish we used Sass. :stuck_out_tongue: Visually this looks great, but I'll leave it to @danrabbit to worry about maintainability concerns.

danrabbit:
The problem with this is that these linked entries don't actually have a border here so this resizes the dialog every time you change focus which looks really janky

danrabbit:
I'm going to close this because of the aforementioned bug

PR #283, opened 2018-03-04
Don't force entry icons to be symbolic

danrabbit:
Assume that developers will do the right thing


cassidyjames:
@danrabbit this looks good and works well in AppCenter. Any other edge cases to look out for?

danrabbit:
This won’t effect any of our apps. It was more trying to be pro-active against third party apps

PR #294, opened 2018-03-17
fix invisible sliders in non-overlay scrollbars

nick87720z:
For now slider background is found only for
 .overlay-indicator scrollbar subclass.

Probably same sitation is in gtk-3.0, as cleanup, made after copying from gtk-3.0, has nothing about removal of similar useful lines.


nick87720z:
I just found true meaning of previous state, before my modification. Probably, scrollbar.overlay-indicator, from which i copied all 3 lines, may be removed.

PR #300, opened 2018-03-22
Scrollbar enhancement

nick87720z:
I made a set of scrollbar enhancement , which appeared to be hard to propose as separate pull requests  (most are too elementary, and it would be too long).
1. Static (non-overlay) scrollbars had artifact in scrollwindow junction, which becomes even more ugly, when scrollbars enlarge on hover.
scrollbar-junction-before
the solution was to replace it with opaque rectangle, which is tiny bit darker, than trough at hovered scrollbar.
2. Static scrollbars in scrolled window had thick 3px border, which was controlled by some hidden setting, which doesn't appear in inspector (however, it may be found in adwaita's gtk-contained.css). Instead, static scrollbars now have light 1px border.
3. Overlay scrollbar trough also got hardening internal border - just an eye candy, for better contrast. Still enough light to perceive scrollbar as content addon.

Screenshots with new static and overlay scrollbars. For static scrollbars don't forget PR #294 :)
nooverlay-scrollbars-new overlay-scrollbars-border

p.s. Should not static scrollbars be a bit thicker, e.g. 5px?

### DIFF
--- a/gtk-3.22/gtk-widgets.css
+++ b/gtk-3.22/gtk-widgets.css
@@ -1753,11 +1753,11 @@ scrollbar.vertical slider {
     min-width: 3px;
 }
 
-scrollbar.overlay-indicator.horizontal:hover slider {
+scrollbar.horizontal:hover slider {
     min-height: 8px;
 }
 
-scrollbar.overlay-indicator.vertical:hover slider {
+scrollbar.vertical:hover slider {
     min-width: 8px;
 }
 
--- a/gtk-3.22/gtk-widgets.css
+++ b/gtk-3.22/gtk-widgets.css
@@ -1816,7 +1816,7 @@ scrollbar.overlay-indicator:hover trough {
 
 * > junction {
     border-style: none;
-    background-color: alpha (#000,0.10);
+    background-color: alpha (#000, 0.1);
 }
 
 /********
--- a/gtk-3.22/gtk-widgets.css
+++ b/gtk-3.22/gtk-widgets.css
@@ -21,7 +21,6 @@
     -GtkTextView-error-underline-color: @error_color;
     -GtkIMHtml-hyperlink-color: @link_color;
     -GtkHTML-link-color: @link_color;
-    -GtkScrolledWindow-scrollbar-spacing: 0;
 
     /* -gtk-icon-palette: needs-attention @attention_color, success @success_color, warning @warning_color, error @error_color; FIXME: Uncomment for 3.19 */
 }
@@ -1799,6 +1798,10 @@ scrollbar.overlay-indicator:hover trough {
 scrolledwindow junction {
     border-style: none;
     background-color: @bg_color;
+}
+
+scrolledwindow {
+    -GtkScrolledWindow-scrollbar-spacing: 0;
 }
 
 /********
--- a/gtk-3.22/gtk-widgets.css
+++ b/gtk-3.22/gtk-widgets.css
@@ -1749,13 +1749,13 @@ scrollbar.horizontal slider {
     min-height: 3px;
 }
 
-scrollbar.horizontal:hover slider {
-    min-height: 8px;
-}
-
 scrollbar.vertical slider {
     min-height: 30px;
     min-width: 3px;
+}
+
+scrollbar.horizontal:hover slider {
+    min-height: 8px;
 }
 
 scrollbar.vertical:hover slider {
--- a/gtk-3.22/gtk-widgets.css
+++ b/gtk-3.22/gtk-widgets.css
@@ -1772,10 +1772,21 @@ scrollbar.overlay-indicator slider:hover {
     border-radius: 20px;
 }
 
-scrollbar.bottom trough { border-top-width: 1px; }
-scrollbar.top trough { border-bottom-width: 1px; }
-scrollbar.left trough { border-right-width: 1px; }
-scrollbar.right trough { border-left-width: 1px; }
+scrollbar.bottom trough {
+    border-top-width: 1px;
+}
+
+scrollbar.top trough {
+    border-bottom-width: 1px;
+}
+
+scrollbar.left trough {
+    border-right-width: 1px;
+}
+
+scrollbar.right trough {
+    border-left-width: 1px;
+}
 
 scrollbar trough {
     padding: 0;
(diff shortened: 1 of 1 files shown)


gtk-3.22/gtk-widgets.css
```diff
@@ -1771,25 +1772,51 @@ scrollbar.overlay-indicator slider:hover {
...
+}
+
+scrollbar.right trough {
+    border-left-width: 1px;
+}
+
 scrollbar trough {
     padding: 0;
+    border-style: solid;
+    border-color: alpha (#000, 0.05);
```
danrabbit: Can we remove this please? I want to keep the trough flat intentionally

danrabbit: Can we remove this please? I want to keep the trough flat intentionally

gtk-3.22/gtk-widgets.css
```diff
@@ -1771,25 +1772,51 @@ scrollbar.overlay-indicator slider:hover {
...
+
+scrollbar:hover trough {
+    background-color: transparent;
+    border-radius: 0;
 }
 
 scrollbar.overlay-indicator trough {
     background-color: transparent;
+    border-style: none;
+    border-color: alpha (#000, 0.15);
```
danrabbit: Same here, no border please

gtk-3.22/gtk-widgets.css
```diff
@@ -1771,25 +1772,51 @@ scrollbar.overlay-indicator slider:hover {
...
 }
 
 scrollbar.overlay-indicator:hover trough {
     background-color: alpha (#000, 0.1);
+    border-style: solid;
```
danrabbit: And here also

gtk-3.22/gtk-widgets.css
```diff
@@ -1771,25 +1772,51 @@ scrollbar.overlay-indicator slider:hover {
...
 }
 
-scrollbar:hover trough {
-    background-color: transparent;
-    border-radius: 0;
-}
-
-trough {
-    background-color: @base_color;
+* > junction {
```
danrabbit: wildcard selector is probably not good. Can we do a specific selector here?

nick87720z: Are there other uses for junction selector, than junction of crossed scrollbars?
I just decided to kill more rabbits... before they appear :)
Of course, it could be scrolledwindow > junction... or it is all possible use case for junction?

nick87720z:
Besides of fancy contrast border, all changes are about non-overlay scrollbars. You could either any gtk3 app from term with GTK_OVERLAY_SCROLLING=0 or selectively toggle overlay type in GtkInspector (Ctrl+Shift+D). However... e.g., Notes-Up somehow doesn't have junction artifact. It just doesn't draw it - probably, it has uses css for that.

For exacts - most of my apps are non-elementary.
qalculate-gtk - official and most featured libqalculate gui (in compareson to NaSC) has junction in History tab.
Sakura and xfce4-terminal force non-overlay type.
Pantheon terminal followes mentioned above env option, but i did not see any junctions in it.
Probably Code/Scratch too (don't know about scratch, as removed it after some fail... don't recall what).
Epiphany... could not open inspector for it. Its scrollbar currently looks like overlay, but doesn't expand on hover (does it just need another selector?)

gtk-3.22/gtk-widgets.css
nick87720z: Hm. If non-overlay scrollbars will have background looking like part of content... will not junction also better to be transparent?

nick87720z: ah... when it is transparent, it has same background, as hovered scrollbar. However, it could be set to base color.

gtk-3.22/gtk-widgets.css
nick87720z: Looks like junction is going to remain. Something locks junction style from change (i tried to place this block at the end of gtk-widgets.css).

gtk-3.22/gtk-widgets.css
```diff
@@ -21,6 +21,7 @@
     -GtkTextView-error-underline-color: @error_color;
     -GtkIMHtml-hyperlink-color: @link_color;
     -GtkHTML-link-color: @link_color;
+    -GtkScrolledWindow-scrollbar-spacing: 0;
```
danrabbit: It looks like instead of using the wildcard selector here that you can do

```
scrolledwindow {
    -GtkScrolledWindow-scrollbar-spacing: 0;
}
```

gtk-3.22/gtk-widgets.css
```diff
@@ -1783,13 +1793,12 @@ scrollbar.overlay-indicator:hover trough {
...
-    background-color: transparent;
-    border-radius: 0;
-}
-
-trough {
-    background-color: @base_color;
+/* FIXME: background is ok, but nothing seems affecting junction border.
+ * Even placing this block at the end of file doesn't help.
+ * Can't find source, unless inspector will show junction node */
+scrolledwindow > junction, {
```
danrabbit: because of the `,` this isn't a valid selector. You probably also don't need the `>` here since we can probably assume there won't be a junction inside of a junction

gtk-3.22/gtk-widgets.css
```diff
@@ -1748,16 +1749,16 @@ scrollbar.horizontal slider {
...
-    min-height: 8px;
-}
-
 scrollbar.vertical slider {
     min-height: 30px;
     min-width: 3px;
 }
 
-scrollbar.vertical:hover slider {
+scrollbar.overlay-indicator.horizontal:hover slider {
```
danrabbit: This means non-overlay scrollbars are always small and can never expand. This would really break click targets here.

gtk-3.22/gtk-widgets.css
nick87720z: yep... just wanted to keep less lines

gtk-3.22/gtk-widgets.css
nick87720z: hm... i thought, expanded non-overlay scrollbars was due to simpler code (selector). When such scrollbar expands, it causes content to resize too. I took example from Adwaita, where expanding was feature of overlay scrollbars, while non-overlay was at full size. However, you then wrote, that they need to be more narrow.

So, should static scrollbars indeed be expanded?
(i also asked for that about to make its width between default and hovered width - i.e. 5 or 6px)

gtk-3.22/gtk-widgets.css
nick87720z: yeah, even adwaita css has it without '>' .

gtk-3.22/gtk-widgets.css
nick87720z: @danrabbit When non-scrollbar expands, its junction produces artifact, described above. But it doesn't happen if scrollbar doesn't change size.

When junction resizes, new border is drawed, bug old area os not repainted with normal background color. When slider is dragged, this somehow causes complete junction redraw with background, but when dragging is completed and scrollbar is unhovered, this produces second sort of artifact.

nick87720z: commited

Issue #299, opened 2018-03-21
regression: some images are broken after pull request #260


nick87720z:
I found, that pull request #260 breaks some things.

Common situation: each image, which is first child of either menuitem (of some kind) or something inside it becomes truncated by 16px.

Two examples (not from elementary os):
- xfce applications menu, generated by garcon with exact structure, when menuitem contains box with icon and label;
- xfce4-pulseaudio-widget: MPRIS playback control buttons, placed in menuitems, representing registered players;
![screenshots-xfappmenu](https://user-images.githubusercontent.com/3896190/37717447-64ff0696-2d42-11e8-99bf-30c0b5e6b4f5.png)
![screenshot-xfpamenu](https://user-images.githubusercontent.com/3896190/37717465-6c9d9a48-2d42-11e8-8c1c-0b54b7b2b0e6.png)

After i wrote to ~/.config/gtk-3.0/gtk.css:
```
image {
    margin: 0px;
}
```
and after finding exact source of trouble:
```
menu:dir(ltr) menuitem image:first-child {
    margin-left: 0px;
}
```
Both places were fixed.

I was close to make PR, with eiher of above solutions (for later - just comment out or remove entire block), but i worry, it would have other side effects.

Are there exact cases, for which PR #260 was made, or it is prophylaxis?

I'm sure, such cases could be solve by another ways.
Simple typing to inspector's css editor:
`* { margin: 0px; }`
Makes it pretty compact. Extend it with `padding: 0px;`, and it will be veeery tiny - menu from first screenshot will look like some themes in old enlightenment, fluxbox, openbox (with some themes).

danrabbit:
I can't reproduce this issue in pantheon anywhere. The selectors are now more specific and should only effect the first child.

PR #302, opened 2018-03-25
Fix issue #299

nick87720z:
After discovering, that PR #260 is reason for issue #299, i looked to its discussion, but could not find any mentions of exact problems, which appeared after menuitem functionality change. While it seems intended to make it more compact - imho it is not good idea to use negative valuds for margin.

p.s. Would be great of course to be able to reference to parrent of some widget class, not just child, as well as enhance selectors with specifier for arbitrary amount of elements in path, which satisfy to specified filter, implemented e.g. like list of selectors.


danrabbit:

PR #330, opened 2018-05-03
Fix specificity issues with ssd decorations

### LINKED ISSUE #329: Strange header behavior
On fully updated Juno daily build I issued with this strange header behavior in some GTK applications, like Deluge, Timeshift, Electrum, Sublime Text 3.

You can see it in action by the link https://youtu.be/EqmtmBiJtIM

danirabbit:
Fixes #329 

You need to restart Gala for changes to take effect


Djaler:
This is works for me

PR #334, opened 2018-05-09
Add an inline toolbar style for tabs

danrabbit:
Adds a style for tabs that can be used with the `inline-toolbar` class and a color variable so you can manually set the color here.

Feedback wanted on the color variable name

screenshot from 2018-05-09 12 16 34 2x
screenshot from 2018-05-09 12 16 49 2x


cassidyjames:
@danrabbit can you fix conflicts?

cassidyjames:
`tab_base_color` sounds sane to me :+1:

PR #340, opened 2018-05-19
Overshoot

cassidyjames:
Experimenting more than anything. I like the undershoot, though it could be refined. The overshoot I'm mixed on; I tried the same color as the undershoot at first which was nice but super subtle. The accent color can be pretty nice, but it might be too bright?


cassidyjames:
The undershoot just seemed too problematic to do globally. I would be interested in revisiting it eventually on a widget-by-widget basis, but for now I've reduced the scope of this PR to just accent-colored overshoot.

Perhaps it's because I've grown so used to it on Android and Pop!_OS, but the accent-colored overshoot seems to provide some nice feedback of "you're at the end of a scroll area" to let you know the OS is still receiving your input, but that you can't go any further. And by using the accent color, branded apps get something that fits in with their brand for free.

@danrabbit try it out for a bit? :smile:

danrabbit:
Wanna resolve conflicts? :)

cassidyjames:
:sob: sure

cassidyjames:
@danrabbit ready for ya

cassidyjames:
@danrabbit using background-size and position means a whole lot more specific sizing for each individual style; I'm okay with the 20% though.

cassidyjames:
@danrabbit ready for you to poke some more. If you think background-size is more clear I'm still open to that. You're also welcome to push to this branch if you want to play with things.

PR #351, opened 2018-07-06
Fix the styling of round image buttons

JosephMcc:


JosephMcc:
Stumbled across the broken round buttons in GtkWidgetFactory on page 3. I'm not sure where else these might show up in elementary.

PR #365, opened 2018-07-25
Style box inside the infobar revealer

danrabbit:
Instead of styling the infobar itself, style the box inside the revealer

Turns out these days infobar has a revealer packed in. By styling the revealer itself, borders are still drawn when the revealer is "closed". By styling the box inside the revealer, borders will be hidden.

Should save on adding revealers for infobars since we can just use the built-in revealed property

Handy infobar reference is either switchboard power plug or gtk3-demo

### DIFF
--- a/gtk-3.22/gtk-widgets.css
+++ b/gtk-3.22/gtk-widgets.css
@@ -3435,7 +3435,7 @@ calendar .view {
 * GtkInfoBar *
 *************/
 
-infobar {
+infobar revealer > box {
     background-color: @bg_color;
     background-image: -gtk-icontheme("dialog-information-symbolic");
     background-size: 16px;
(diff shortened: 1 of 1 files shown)


ztefn:
Searchbars could use a similar treatment, see for example the GtkSearchBar inside this GtkShortCutsWindow below. When closed there's an extra thick border below the headerbar caused by the visible border from the searchbar.
screenshot from 2018-07-31 22 29 55
screenshot from 2018-07-31 22 29 45

danrabbit:
@ztefn can you open a new issue report for this please? :)

ztefn:
@danrabbit sure, although none of [my issues](https://github.com/elementary/stylesheet/issues/created_by/ztefn) ever got any response other than from myself...

PR #372, opened 2018-08-08
Fix dark infobar buttons

danrabbit:
Update some selectors and rules so that buttons in infobars look right in the dark stylesheet


cassidyjames:
Nice catch.

PR #373, opened 2018-08-08
Remove unused dark spinbutton rules

danrabbit:
These selectors are no longer valid and it seems that spinbuttons already look right in the dark stylesheet


cassidyjames:
Otherwise the removal here looks right to me.

danrabbit:
Oops, you're right. Also invalid selectors. Nice catch :)

PR #376, opened 2018-08-13
Slate Variant

cassidyjames:
This moves the main stylesheet to an `elementary` subdirectory, and adds an `elementary-slate` directory next to it that is mostly symlinks to the main stylesheet, with just a different `gtk.css` and `gtk-dark.css` set of variables.

Fixes #375 

# Todo:

- [x] Draw back button assets
- [x] Figure out Plank badge (#545)

PR #377, opened 2018-08-13
Add slate variant to packaging

cassidyjames:
Depends on #376


danrabbit:
Closing because greenfield

PR #386, opened 2018-08-15
Add a WIP high contrast style

cassidyjames:
Something I was playing with. Putting here to document it more than anything.

image 1


cassidyjames:
Note, we'd want to expose this as a "increase contrast" setting or something, as it's likely not going to be as drastic as other high contrast modes, especially to start.

danrabbit:
Closing because Greenfield

PR #388, opened 2018-08-16
Split out brand files

cassidyjames:
Another small split out from #376. This makes it more clear what are supported brand modifications, plus makes color variants even easier. :wink:


cassidyjames:
@danrabbit whoops. Fixed.

PR #393, opened 2018-09-08
Remove overly specific style

cassidyjames:
We weren't sure what this was for, and it's causing weird styling in the AppCenter StripeDialog.


danrabbit:
weird

PR #402, opened 2018-09-28
Add class for accelerator

danrabbit:
Turns out this is a Gtk style class constant


Alecaddd:
Tested and it works for me!

PR #405, opened 2018-09-30
Soften gradient on Gtk.TreeView headers so it looks good with inline tabs

### LINKED ISSUE #383: Treeview column headers don't match inline tabs
Treeview headers should be updated to blend in with inline tabs better

screenshot from 2018-08-14 16 07 29 2x

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/62147061-treeview-column-headers-don-t-match-inline-tabs?utm_campaign=plugin&utm_content=tracker%2F45189256&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F45189256&utm_medium=issues&utm_source=github).
</bountysource-plugin>

lainsce:
Makes it so the treeview column header widgets not clash with inline tabs, by removing the old gradient.
Also works with Dark theme.
Fixes #383.

Old:
old-treeview-headers

New:
new-treeview-headers

### DIFF
--- a/elementary/gtk-3.0/gtk-widgets-dark.css
+++ b/elementary/gtk-3.0/gtk-widgets-dark.css
@@ -684,40 +684,15 @@ GtkAssistant .sidebar {
  * Column Headers *
  *********************/
 column-header .button {
-    background-image:
-        linear-gradient(
-            to bottom,
-            shade (
-                @titlebar_color,
-                0.88
-            ),
-            @titlebar_color
-        );
-    border-image:
-        linear-gradient(
-            to bottom,
-            shade (
-                @titlebar_color,
-                0.88
-            ),
-            shade (
-                @titlebar_color,
-                0.7
-            )
-        ) 0 0 1 1;
+    background-image: none;
+    background-color: shade (@titlebar_color, 0.88);
+    border-color: shade (@titlebar_color, 0.88);
+    border-image: none;
     font-weight: bold;
 }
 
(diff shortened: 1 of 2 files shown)


cassidyjames:
I'm pretty hesitant on this because the previous style had some nice depth, and the new style goes super flat everywhere for one specific use case that we're not even doing yet in one app.

lainsce:
I could reverse the gradient, and make it a tad less flat but also not a dark gradient. How's that sound?

cassidyjames:
@lainsce I'd be curious to see, but I actually like the current inset style.

lainsce:
Something like this:
screenshot from 2018-10-01 14 57 29

cassidyjames:
@lainsce my other worry is that this makes the tabs look visually attached to the column headers only, and not the rest of the tab contents. Honestly I am considering that the inline style should just exist on the icon view where there's an unbroken association between the tab and the contents.

lainsce:
It could be that when it's not in iconview in files, it loses the inline tab style instead of this PR. Should I close the PR then?

cassidyjames:
@lainsce I'd be interested in hearing more feedback from @elementary/ux peeps as well, but that's my personal opinion. :)

lainsce:
Going to close this one since it affects too much the other places where the gradient helps.

PR #409, opened 2018-10-01
Fix menuitem color from suggested-action buttons

danrabbit:
These selectors may have needed to be this specific in an older version of the stylesheet, but I don't see why they need to be anymore. Tested this with suggested action buttons in headers in dark and light with brand colors.

## BEFORE
screenshot from 2018-10-01 15 01 02 2x

## AFTER
screenshot from 2018-10-01 15 15 15 2x

### DIFF
--- a/elementary/gtk-3.0/gtk-widgets.css
+++ b/elementary/gtk-3.0/gtk-widgets.css
@@ -2616,14 +2616,9 @@ button.suggested-action,
         0 1px 0 0 alpha (#fff, 0.15);
 }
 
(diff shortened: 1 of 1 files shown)


cassidyjames:
I'm running this locally and haven't seen any ill-effects. Is there a branch of the screenshot tool (or another app) somewhere where I can test the fix?

danrabbit:
@cassidyjames I didn't save that prototype unfortunately. You could use Simple Scan and add the `suggested-action` class to the menu button in the header there

PR #410, opened 2018-10-01
Unblur the icon used to show Infobar Type

lainsce:
This makes it so the icon doesn't land in a half-pixel, making it not blur.

Looks like this:
Old =
screenshot from 2018-10-01 19-43-24

New =
screenshot from 2018-10-01 19-43-55


danrabbit:
Can you try with 1em instead of 12px? I think that should still give us an integer value but also adapt better with the large fonts a11y option

PR #419, opened 2018-10-10
Fix separator margins of menus created with gtk_menu_button_set_menu_model()

rkoesters:
The spacing between section separators in popover menus created with [gtk_menu_button_set_menu_model()](https://developer.gnome.org/gtk3/stable/GtkMenuButton.html#gtk-menu-button-set-menu-model) does not have the same spacing as other menu separators. This pull request attempts to fix that.

Popover menus created this way still have margin properties (of GtkPopover -> GtkStack -> GtkMenuSectionBox) set to 10, which creates spacing around the menu itself that isn't consistent with elementary's other menus, however I don't know how to fix this in the stylesheet. [My app fixes it by setting those properties in the code](https://github.com/rkoesters/xkcd-gtk/blob/master/cmd/xkcd-gtk/style.go#L128).


rkoesters:
Examples from my app:

Before patch (without changing margin properties):
popover-menu-without-patch-margin-prop-10

After patch (without changing margin properties):
popover-menu-with-patch-margin-prop-10

Before patch (with changing margin properties):
popover-menu-without-patch-margin-prop-4-0

After patch (with changing margin properties):
popover-menu-with-patch-margin-prop-4-0

danrabbit:
Is there a more specific selector for a separator in a model menu? I feel like this is going to cause unintended results for popovers where we don't want to have extra margin on a separator always, such as when it's butted up against a scrollbox or something

rkoesters:
Yeah, that's a valid concern. I did test the original patch with every default elementary app in addition to a couple GNOME apps (dconf-editor, gnome-disks and epiphany) and it didn't affect any of them in negative ways (my first iteration, not shown here, did affect them, so I tweaked it until it didn't).

Here is the object hierarchy:
hierarchy

And here is the CSS node hierarchy:
css-nodes

Unfortunately, it doesn't look like there is a way to select the GtkMenuSectionBox specifically (like `menusectionbox` or something), but I did make the selector more specific.

rkoesters:
Note: this affects Evince (it uses `gtk_menu_button_set_menu_model()` for its menus)

danrabbit:

PR #433, opened 2018-10-27
Fix headerbar buttons focused text color in dark theme.

rkoesters:
If the dark theme is enabled, the focused text color of buttons in the titlebar doesn't use textColorPrimary, it uses a shade of white instead. This pull request tries to fix that so we don't have weird contrast in the titlebar buttons.

Before patch:
screenshot from 2018-10-27 13 14 07

After patch:
screenshot from 2018-10-27 13 14 32


rkoesters:
I tried to put this addition where it belongs, but it is unclear to me whether it should go under `Buttons`, `CSD`, or `Header Bars`.

danrabbit:
OP, what application are you reproducing this issue in? I'm not sure this looks right in either of your before or after shots. Are you saying this is a problem that occurs when you are manually overriding the system default to be the dark stylesheet or are you developing an application that uses the `prefer-dark-theme` setting?

danrabbit:

PR #443, opened 2018-11-09
Add accent support to scales

cassidyjames:
Like #262 but for scales. i.e. so we can do stuff like this:

screenshot from 2018-11-08 18 15 10

### DIFF
--- a/elementary/gtk-3.0/gtk-widgets.css
+++ b/elementary/gtk-3.0/gtk-widgets.css
@@ -847,7 +847,11 @@ scale.vertical mark indicator {
 }
 
 scale highlight {
-    background-color: alpha (#000, 0.45);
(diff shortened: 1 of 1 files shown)


cassidyjames:
Also allows interesting app branding like this

image 11

harisvsulaiman:
This is very good for developers who brand apps.
I think the whole widget set should handle branding effectively.

cassidyjames:
@harisvsulaiman yeah that's the idea. Currently anything that is accent colored (i.e. blue by default) does already follow the app's accent color, like switches, focus outlines, selection colors, overscroll feedback, check boxes, radios, etc. The difference here is that scales were not accent colored before, so this adds a class to enable that.

Issue #445, opened 2018-11-10
Color active icon of ModeSwitch


cassidyjames:
Since we lost the accent color from the normal switch, it could be interesting to color the active icon in ModeSwitches instead.

cassidyjames:
![image 14](https://user-images.githubusercontent.com/611168/48295052-aeb87280-e445-11e8-83b6-956f05bb8984.png)

![image 13](https://user-images.githubusercontent.com/611168/48295057-b2e49000-e445-11e8-88c6-4faaf10e9ed9.png)

PR #446, opened 2018-11-10
Support directly-styled titlebar images and labels

cassidyjames:
So we can support #445 in headerbars


danrabbit:

Issue #440, opened 2018-11-03
Switches are low-contrast in dark style


cassidyjames:
Switches in the dark style are very low contrast and are pretty hard to see at both standard and mode-switch sizes.

![image](https://user-images.githubusercontent.com/611168/47945700-50c9df00-dec9-11e8-8f38-52630532f075.png)

micahilbery:
![screenshot from 2018-11-02 22 55 16](https://user-images.githubusercontent.com/10608836/47947842-7e774e00-def2-11e8-861a-d81ee13685dc.png)

Maybe something similar to what we did with scales in the dark style would work here as well giving it more of a silver slider and background.

cassidyjames:
@micahilbery that direction looks sane to me. It's much easier to see at a glance!

PR #449, opened 2018-11-20
Always use light switch sliders

cassidyjames:
This translates the light style sliders over to the dark style, much like we did for scales. This makes them a lot easier to spot and higher contrast.

| Before | After |
|-----|-----|
| screenshot from 2018-11-20 15 03 21 | screenshot from 2018-11-20 15 30 24 |
| screenshot from 2018-11-20 15 03 16 | screenshot from 2018-11-20 15 30 19 |

### DIFF
--- a/elementary/gtk-3.0/gtk-widgets.css
+++ b/elementary/gtk-3.0/gtk-widgets.css
@@ -2208,13 +2208,14 @@ switch slider {
             shade(@SILVER_100, 1.05),
             shade(@SILVER_100, 0.95)
         );
+    border: 1px solid alpha(#000, 0.2);
     border-radius: 50%;
     box-shadow:
         inset 0 0 0 1px alpha(#fff, 0.05),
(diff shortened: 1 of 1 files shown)


lainsce:
Maybe tone down it a bit, the color right now is too white imo.

peteruithoven:
Another way to tackle this might be to have the blue all around the switch, like a border. This currently only happens in the light theme.

cassidyjames:
@lainsce I tweaked it a bit now, but I am matching the new contrasty scales exactly.

cassidyjames:
Also, still WIP ;)

@peteruithoven ah I see what was going on there. Yeah I think we'll want to keep that across styles.

cassidyjames:
Another example: here's a modeswitch.

screenshot from 2018-11-20 15 35 05
screenshot from 2018-11-20 15 35 01

lainsce:
@cassidyjames now it looks better :3

cassidyjames:
Fixed #440

PR #452, opened 2018-12-02
Remove the rounded edges of tabs

satran:
I would like to make this a proposal for removing the rounded edges for the tabs. Have a look at the current style of the tabs in the Terminal application.
screenshot from 2018-12-02 11 34 30
The problem I see with this is that it is difficult to say where the tab belongs to. Using the round edges it seems to belong to the body. But the colour of the tab is different from that of the body. Code does a good job in integrating it with the color of the body. Basing on the colour of the tabs I would assume it belongs to the header bar/ title bar. Then the rounded corners of the tab is very misleading. The PR tries to solve this by removing the rounded edges. Here is how it looks now:
screenshot from 2018-12-02 11 34 57

### DIFF
--- a/elementary/gtk-3.0/gtk-widgets.css
+++ b/elementary/gtk-3.0/gtk-widgets.css
@@ -1676,7 +1676,7 @@ notebook tab:checked {
         );
     border-color: shade (@titlebar_color, 0.62);
     border-image: none;
-    border-radius: 4px 4px 0 0;
+    border-radius: 0;
     border-style: solid;
     box-shadow:
         inset 0 0 0 1px alpha (@bg_highlight_color, 0.2),


cassidyjames:
My concern with this style is that it loses a strong signifier for which is the active tab; instead of having two signifiers—depth and shape—it is reduced to just one—depth.

danrabbit:
I agree with Cassidy. I'm in favor of closing this as "Design Conflict" since it makes the active tab harder to recognize

satran:
I doubt clarity of which is the active tab is a valid concern. If you look at the screenshot below I doubt one would have difficulty in distinguishing an active tab.
screenshot from 2018-12-18 20 38 18

And the rounded edges are so small sometimes it is not very visible. I'm not sure it is a strong signifier. If the extra signifier is needed I would suggest going the firefox approach or even GNOME's like the following screenshot.
screenshot from 2018-12-18 20 42 35

Anyways I just found the rounded edges a bit misplaced. Elementary is a well designed OS I hope you find a good alternate for the tabs.

PR #457, opened 2018-12-16
Update .seek-bar scale slider style.

Mohelm97:
Fixes #301
Seek bar slider should be zero in width and height, it's not shown, but if there's width or height the seek bar will not exactly go where you clicked.

Before:
screen record from 2018-12-16 08 26 50

After:
screen record from 2018-12-16 08 25 37


peteruithoven:
Might also fix: https://github.com/elementary/videos/issues/103

Mohelm97:
@peteruithoven
Yeah it does for any thing uses the seek bar

PR #461, opened 2018-12-28
Shade switches with white overlay

danrabbit:
Turns out that shade works okay for dark colors like blue, but looks like garbage for green. Instead, overlay white:

screenshot from 2018-12-27 18 40 45 2x

screenshot from 2018-12-27 18 41 44 2x


danrabbit:
screenshot from 2018-12-28 10 09 22 2x

PR #462, opened 2019-01-03
Styles for new prototype avatar

danrabbit:
screenshot from 2019-01-03 11 45 15 2x


elementary/gtk-3.0/granite-widgets.css
```diff
@@ -23,6 +23,21 @@
 * Avatar *
 *********/
 
+avatar {
+    color: @SILVER_900;
+    font-weight: 600;
+}
+
+avatar label {
+    background-color: @SILVER_300;
```
cassidyjames: Would we want to use white-on-blue for now to be more similar to the placeholder face? Ideally eventually I'd like to do a color based on the name hash, maybe from 20 palette colors (light and dark version of each color).

danrabbit: tbh I think it doesn't look as good:

screenshot from 2019-01-03 14 13 34 2x

cassidyjames: @danrabbit the borders and shadows don't look right there. But I'm also mostly indifferent about the color, especially if we plan to do something with programmatic colors down the line. :smile:

4jNsY6fCVqZv:
How does a little more colour play together with your design concept?
This seems to be the GNOME prototype: https://gitlab.gnome.org/Teams/Design/os-mockups/-/commit/f9bd13fee5bdcb36db481ca63a7ff23cc6d780ed

PR #470, opened 2019-02-02
Use shadow on titlebar instead of border

danrabbit:
Gives us much sharper borders when listboxes are up against the titlebar or (for example) in AppCenter:

BEFORE:
screenshot from 2019-02-01 21 23 16 2x

AFTER:
screenshot from 2019-02-01 21 23 01 2x

Needs lots of testing for regressions


cassidyjames:
Giving this a go again for a while to test, but I haven't seen anything out of place so far..

cassidyjames:
One thing I've noticed: in non-native apps like Firefox and Slack, there's a new light-colored bottom border on the titlebar.

Firefox

Slack

cassidyjames:
@danrabbit any chance to revisit this? It does look a lot better when it works.

danrabbit:
@cassidyjames I'm kinda feeling like we should save it for greenfield now

PR #479, opened 2019-02-15
Invert sidebar badge colors

danrabbit:
## BEFORE
screenshot from 2019-02-14 20 40 57 2x

## AFTER
screenshot from 2019-02-14 20 40 37 2x


tintou:
Also I don't know if you fixed it, but you might look at the window-unfocused issue with the badge

cassidyjames:
@tintou it seems fixed (or at least less-bad) in this branch.

PR #490, opened 2019-04-05
Add support for progressbar.osd

### LINKED ISSUE #489: Issue with GtkProgressBar
Here is a tiny Vala snippet:
```vala
public class MyApplication : Gtk.Application {
    public MyApplication () {
        Object(application_id: "testing.my.application",
                flags: ApplicationFlags.FLAGS_NONE);
    }

    protected override void activate () {
        // Create the window of this application and show it
        Gtk.ApplicationWindow window = new Gtk.ApplicationWindow (this);
        var headerbar = new Gtk.HeaderBar ();
        var entry = new Gtk.Entry ();
        var progressbar = new Gtk.ProgressBar ();
        progressbar.get_style_context ().add_class ("osd");
        progressbar.valign = Gtk.Align.END;
        progressbar.fraction = 0.5f;
        var overlay = new Gtk.Overlay ();

        overlay.add (entry);
        overlay.add_overlay (progressbar);
        headerbar.set_custom_title (overlay);
        window.set_titlebar (headerbar);
        window.show_all ();
    }

    public static int main (string[] args) {
        MyApplication app = new MyApplication ();
        return app.run (args);
    }
}
```
run it with `vala ./test.vala --pkg gtk+-3.0`

How it should look like:
Capture d'écran du 2019-04-05 12 12 09@2x

How it currently looks like:
Capture d'écran du 2019-04-05 12 13 04@2x

danrabbit:
Fixes #489

### DIFF
--- a/elementary/gtk-3.0/gtk-widgets.css
+++ b/elementary/gtk-3.0/gtk-widgets.css
@@ -2787,19 +2787,25 @@ button.destructive-action:focus,
 
 button.flat.destructive-action,
 button.flat.destructive-action:backdrop,
-button.flat.destructive-action:hover {
+button.flat.destructive-action:hover,
+actionbar button.image-button.destructive-action,
+actionbar button.image-button.destructive-action:backdrop,
+actionbar button.image-button.destructive-action:hover {
     background: none;
     border: none;
     box-shadow: none;
 }
 
-button.flat.destructive-action:focus {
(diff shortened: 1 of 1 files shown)


tintou:
I think that it's too big for the input entry.
To see it in action in Builder `GTK_DEBUG=interactive flatpak run org.gnome.Builder`

PR #495, opened 2019-04-13
Add `.accent` support to buttons

### LINKED ISSUE #494: Add accent support to buttons
It'd be handy to be able to add `.accent` to buttons to get all the required styles for suggested-action-like buttons but with the accent color instead.

cassidyjames:
Fixes #494


danrabbit:
Gonna close this since we've decided for greenfield that suggested action buttons will follow the accent color so it would be the same thing in the end

PR #500, opened 2019-05-10
Adjust Granite.StorageBar when using dark style

### LINKED ISSUE #498: Adjust StorageBar background when using dark theme
| Light | Dark |
|-------|------|
| Screenshot von 2019-05-09 19 37 33 | Screenshot von 2019-05-09 19 37 22 |

It currently looks like this and imho the background should be also dark when using a dark theme :)

hannesschulze:
Fixes #498 (I don't really know what's the best way to test the installer styles and if the installer actually needs to support dark styles)

| Before | After |
| --- | --- |
| Screenshot von 2019-05-10 16 41 16@2x | Screenshot von 2019-05-10 16 47 03@2x |

### DIFF
--- a/elementary/gtk-3.0/granite-widgets-dark.css
+++ b/elementary/gtk-3.0/granite-widgets-dark.css
@@ -115,24 +115,16 @@
         );
 }
 
-.storage-bar .fill-block.app {
-    border-color: shade (@GRAPE_900, 0.95);
-}
-
 .storage-bar .fill-block.audio {
-    border-color: shade(@ORANGE_900, 0.85);
+    border-color: alpha (@ORANGE_900, 0.65);
 }
 
 .storage-bar .fill-block.files {
-    border-color: shade(@BLUEBERRY_900, 0.85);
+    border-color: alpha (@BLUEBERRY_900, 0.65);
 }
 
 .storage-bar .fill-block.photo {
-    border-color: shade(@BANANA_900, 0.7);
-}
-
-.storage-bar .fill-block.video {
-    border-color: shade(@STRAWBERRY_900, 0.85);
+    border-color: alpha (@BANANA_900, 0.4);
 }
 
 /***************
--- a/elementary/gtk-3.0/granite-widgets-dark.css
+++ b/elementary/gtk-3.0/granite-widgets-dark.css
@@ -88,6 +88,33 @@
     box-shadow: none;
 }
 
+/**************
+* Storage Bar *
+**************/
+
+.storage-bar .fill-block {
+    background-image:
+        linear-gradient(
+            to bottom,
+            @BLACK_300,
+            @BLACK_500
+        );
+    box-shadow:
+        inset 0 0 0 1px alpha (#ccc, 0.05),
+        inset 0 1px 0 0 alpha (#ccc, 0.15),
+        inset 0 -1px 0 0 alpha (#ccc, 0.05);
+}
+
+.storage-bar .fill-block.empty-block,
+.storage-bar .fill-block.unused {
+    background-image:
+        linear-gradient(
+            to bottom,
+            mix(@BLACK_100, @BLACK_300, 0.5),
+            @BLACK_300
+        );
+}
+
 /***************
 * Fancy Scales *
 ***************/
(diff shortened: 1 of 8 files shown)


cassidyjames:
@hannesschulze nice! I like the end result, but I have a thought: would it work to use the same styling in both the light and dark variants, i.e. by using a shade of gray and alpha to get the desired look? If not that's also cool, and we can maintain the styles in both the standard stylesheet and the dark one.

hannesschulze:
I played around with it a bit - at least the box-shadow would cause some problems and would have to be specified separately, and also I only got the background to look a bit too dark in the light version - this could probably be fixed with a little bit of fine-tuning. But imho it wouldn't really be worth it :)

elementary/gtk-3.0/granite-widgets-dark.css
```diff
@@ -88,6 +88,45 @@
...
+
+.storage-bar .fill-block {
+    background-image:
+        linear-gradient(
+            to bottom,
+            @BLACK_300,
+            @BLACK_500
+        );
+    box-shadow:
+        inset 0 0 0 1px alpha (#ccc, 0.05),
```
cassidyjames: In any new stylesheet code, we've been omitting the space. Code syntax highlights properly without the space, plus it's more consistent with built-in CSS methods and languages like Sass.

```suggestion
        inset 0 0 0 1px alpha(#ccc, 0.05),
```

elementary/gtk-3.0/granite-widgets-dark.css
```diff
@@ -88,6 +88,45 @@
...
+        inset 0 1px 0 0 alpha (#ccc, 0.15),
```
cassidyjames:
```suggestion
        inset 0 1px 0 0 alpha(#ccc, 0.15),
```

elementary/gtk-3.0/granite-widgets-dark.css
```diff
@@ -88,6 +88,45 @@
...
+        inset 0 -1px 0 0 alpha (#ccc, 0.05);
```
cassidyjames:
```suggestion
        inset 0 -1px 0 0 alpha(#ccc, 0.05);
```

elementary/gtk-3.0/granite-widgets-dark.css
```diff
@@ -88,6 +88,45 @@
...
+    background-image:
+        linear-gradient(
+            to bottom,
+            mix(@BLACK_100, @BLACK_300, 0.5),
+            @BLACK_300
+        );
+}
+
+.storage-bar .fill-block.audio {
+    border-color: alpha (@ORANGE_900, 0.65);
```
cassidyjames:
```suggestion
    border-color: alpha(@ORANGE_900, 0.65);
```

elementary/gtk-3.0/granite-widgets-dark.css
```diff
@@ -88,6 +88,45 @@
...
+}
+
+.storage-bar .fill-block.files {
+    border-color: alpha (@BLUEBERRY_900, 0.65);
```
cassidyjames:
```suggestion
    border-color: alpha(@BLUEBERRY_900, 0.65);
```

elementary/gtk-3.0/granite-widgets-dark.css
```diff
@@ -88,6 +88,45 @@
...
+}
+
+.storage-bar .fill-block.photo {
+    border-color: alpha (@BANANA_900, 0.4);
```
cassidyjames:
```suggestion
    border-color: alpha(@BANANA_900, 0.4);
```

elementary/gtk-3.0/granite-widgets-dark.css
hannesschulze: I'll go ahead and fix that, but this is also the case in https://github.com/elementary/stylesheet/blob/master/elementary/gtk-3.0/granite-widgets.css#L274 where I copied this from - Should I fix this here too or make a new PR for it?

danrabbit:
Some of the border changes look much better,  but I'm not sure about changing the color of the empty fill block. I wonder if the higher contrast here is better

cassidyjames:
@danrabbit changing it makes sense to me since it's "empty" and thus in the light style it matches the window background. The "other" category might be a bit low contrast, though. Slate might be a better color for that in both styles, anyway.

PR #510, opened 2019-05-25
Use Black 500 from the official palette for the Dark style

lainsce:
As the title says, use Black 500 instead of an untracked blue-ish black not from the palette.

Screenshot of new black color:
Screenshot from 2019-05-24 21 50 32

### DIFF
--- a/elementary/gtk-3.0/gtk-dark.css
+++ b/elementary/gtk-3.0/gtk-dark.css
@@ -20,7 +20,7 @@
 @import url("brand-dark.css");
 
 /* Default color scheme */
-@define-color base_color #3d4248;
+@define-color base_color @BLACK_500;
 @define-color bg_color shade (@base_color, 0.96);
 @define-color bg_highlight_color shade (@bg_color, 1.4);
 @define-color border_color alpha (@BLACK_900, 0.25);
--- a/elementary/gtk-3.0/gtk-dark.css
+++ b/elementary/gtk-3.0/gtk-dark.css
@@ -20,7 +20,7 @@
 @import url("brand-dark.css");
 
 /* Default color scheme */
-@define-color base_color @BLACK_500;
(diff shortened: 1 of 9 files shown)


cassidyjames:
Current dark style in the same context

image

elementary/gtk-3.0/gtk-dark.css
```diff
@@ -20,7 +20,7 @@
 @import url("brand-dark.css");
 
 /* Default color scheme */
-@define-color base_color #3d4248;
+@define-color base_color @BLACK_500;
```
cassidyjames: This feels _quite_ a bit darker which might cause issues. It looks like something in this range is closer to the current brightness of the dark style:

```suggestion
@define-color base_color mix(@BLACK_500, @BLACK_300, 0.5);
```

lainsce: I like that mix() combo. I'll add it to the PR~

micahilbery:
I really like the cooler color of the current styles but I also like using palette variable colors where possible. We could mix `SLATE` with `BLACK` to achieve that if I'm not alone in liking the cooler color.
Dark_Styles_Slate_Mix

lainsce:
The problem with mixing with `SLATE`, is that then the Slate accent option from the other PR gets lost and blends in.

cassidyjames:
Yeah, part of the reason was to move the stylesheet to be more color neutral; if it's blue-tinted, then color-sensitive work i.e. in Photos or a video editor can be compromised. So far I'm a fan of the more neutral dark style.

lainsce:
With that approved review, I merge this!

cassidyjames:
Screenshots for further comparison:

## Cool
Screenshot from 2019-08-04 22 19 12@2x

## Neutral 

Screenshot from 2019-08-04 22 17 23@2x

PR #512, opened 2019-05-30
Fix arrow direction in notebook header

danrabbit:


peteruithoven:
## Before 
Screenshot from 2019-07-01 17-18-05
## After
Screenshot from 2019-07-01 17-16-28

PR #515, opened 2019-06-04
Add AppData

danrabbit:


tintou:
I haven't found any documentation for `type=runtime`, it seems to me that we have to solve https://github.com/ximion/appstream/issues/67 before going forward

danrabbit:
Yeah, I found the `type=runtime` thing in someone else' flatpak repo for the stylesheet. Wasn't sure if it was a real thing or not

tcitworld:
AppStream version `0.12.8` has support for the `icon-theme` component type.

PR #519, opened 2019-06-24
Add support for type-heading class

### LINKED ISSUE #518: Add new Gtk text style classes
It sounds like Gtk4 will have style class constants that replace our `h(1-4)` classes. https://gitlab.gnome.org/GNOME/gtk/issues/1808

We should support these style class for better compatibility with future Gtk apps

Adwaita master now how has:
```
.large-title {
   font-weight: 300;
   font-size: 24pt;
   letter-spacing: 0.2rem;
}
.title-1 {
   font-weight: 800;
   font-size: 20pt;
}
.title-2 {
   font-weight: 800;
   font-size: 15pt;
}
.title-3 {
   font-weight: 700;
   font-size: 15pt;
}
.title-4 {
   font-weight: 700;
   font-size: 13pt;
}
.heading {
   font-weight: 700;
   font-size: 11pt;
}
.body {
   font-weight: 400;
   font-size: 11pt;
}
.caption-heading {
   font-weight: 700;
   font-size: 9pt;
}
.caption {
   font-weight: 400;
   font-size: 9pt;
}
```

danrabbit:
Fixes #518

### DIFF
--- a/elementary/gtk-3.0/granite-widgets.css
+++ b/elementary/gtk-3.0/granite-widgets.css
@@ -619,18 +619,21 @@ window.rounded decoration {
 }
 
 .h4,
-.category-label {
+.category-label,
+.type-heading {
     color: alpha (@text_color, 0.7);
     font-weight: bold;
     text-shadow: 0 1px @text_shadow_color;
 }
 
-.h4 {
+.h4,
+.type-heading {
     padding-bottom: 6px;
     padding-top: 6px;
 }
 
-list .h4 {
+list .h4,
+list .type-heading {
     padding-left: 6px;
 }
 


danrabbit:
Does this solve the problem for you @Kekun?

Kekun:
I am not *exactly* affected by this, but two things to keep in mind:
- https://gitlab.gnome.org/GNOME/gtk/issues/1808 isn't fixed yet and it may still change, I recommend you to wait a bit or to discuss what to expect there so you don't have to fix it later, especially as no app use them yet so there is no hurry;
- these styles will likely be in GTK 4, but not GTK 3 as it is frozen.

I otherwise think it's the right direction to take. :)

PR #521, opened 2019-06-27
Make Entry and SpinButton Progess more subtle

danrabbit:
A little less obnoxious:

Screenshot from 2019-06-27 12 37 22@2x

### DIFF
--- a/elementary/gtk-3.0/apps.css
+++ b/elementary/gtk-3.0/apps.css
@@ -62,6 +62,24 @@
         );
 }
 
+.floating-bar {
+    background-color: alpha (#222, 0.9);
+    border-radius: 3px;
+    border-width: 0;
+    box-shadow:
+        0 1px 3px alpha (#000, 0.12),
+        0 1px 2px alpha (#000, 0.24);
+    color: #fff;
+    padding: 2px 0;
+    margin: 6px;
+    text-shadow: 0 1px 2px alpha (#000, 0.6);
+}
+
+.floating-bar label {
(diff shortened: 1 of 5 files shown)


cassidyjames:
Hmm, I'm mixed. I think it makes sense to move it to the bottom, but this feels out of place. Would it be possible to do like a bottom glow of some sort? I'll try and look for an example of what I mean or mock something up.

cassidyjames:
@danrabbit is something like this too subtle?

image

danrabbit:
Hmm, maybe going the other direction? It seems to kind of conflict with the inset perspective of the entry

cassidyjames:
@danrabbit hm, you lose the fuzziness on the edges which doesn't look quite as good, but I'm fine with the concept.

image

```css
    box-shadow: inset 0 -7px 3px -6px @colorAccent;
    margin: 0 -3px -4px;
```

…wait, we could probably do this with a background gradient.

cassidyjames:
Background gradient style. Still not super happy with the sharp edges, though. But I think it's close.

image

Just remove the border and change the background-image to:

```css
    background-image: linear-gradient(alpha(@base_color, 0) 75%, @colorAccent);
```

danrabbit:
Screenshot from 2019-06-27 14 43 51@2x

How about this?

cassidyjames:
@danrabbit I'm not sure if the border radius on the gradient is better or worse than a hard edge. It looks a little out of place to me, though.

PR #522, opened 2019-07-02
Add undershoot to popovers

cassidyjames:
Undershoots help smooth out the edges of scrollable widgets. We opted to look at it on a widget-by-widget basis in #340, so here's the start of that for popovers.

## Before

Notice the hard edges on the user list.

Before

## After

Soft edges on the user list.

After

### DIFF
--- a/elementary/gtk-3.0/gtk-widgets.css
+++ b/elementary/gtk-3.0/gtk-widgets.css
@@ -3754,6 +3754,22 @@ popover.menu,
     background-image: none;
 }
 
+popover undershoot.top {
+    background:
+        linear-gradient(
+            @bg_color 0%,
+            alpha(@bg_color, 0) 50%
+        );
+}
+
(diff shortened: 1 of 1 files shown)


cassidyjames:
Another place to look is the Wi-Fi menu in the network indicator

peteruithoven:
@cassidyjames any change you could crop those images?

lainsce:
I don't think cropping photos would solve the problem. Any type of image would be used as user photo, even logos.

peteruithoven:
Sorry, to clarify I was refering to the screenshots, to make the examples easier to see. I wasn't suggesting an alternative solution.

peteruithoven:
## Before
Before
## After
After

cassidyjames:
A couple of more screenshots showing the change:

| Bottom cut off | Both ends cut off |
|---|---|
| wifi-1 | wifi-2 |

PR #524, opened 2019-07-08
Style keycaps in menus

danrabbit:
Screenshot from 2019-07-08 16 40 50@2x
Screenshot from 2019-07-08 16 43 16@2x

Passes WCAG AA

### DIFF
--- a/elementary/gtk-3.0/gtk-widgets-dark.css
+++ b/elementary/gtk-3.0/gtk-widgets-dark.css
@@ -293,7 +293,8 @@ switch:checked slider {
         inset 0 -1px 0 0 alpha (#fff, 0.01);
 }
 
-menu .keycap {
+menuitem .keycap,
+modelbutton .keycap {
     background: alpha (#000, 0.15);
     box-shadow: 0 1px 0 0 alpha (#000, 0.25);
     color: alpha(@text_color, 0.75);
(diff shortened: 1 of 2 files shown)


peteruithoven:
I love that we're adding keyboard shortcuts to menu's but with this styling aren't the menu's going to be a bit to busy? Aren't we giving them to much visual priority?

danrabbit:
@peteruithoven there's some conversation here for context: https://twitter.com/DanielFore/status/1148359052532969473

peteruithoven:
Ah I see there are also more versions here: https://github.com/elementary/granite/pull/300

But what about not fulling the buttons with darker gray and using just a dark gray background? That way you could make the text a lighter (less visually prominent) gray as well. 

I don't see an easy way to quickly play with this myself.

peteruithoven:
I'm thinking of larger menu's like the one in Files:
Screenshot from 2019-07-09 02 09 16@2x

danrabbit:
Something I want to look at also is maybe making these shorter so that menu items are less grid-y and more row-y

danrabbit:
Okay latest commit makes the light style much more subtle. Everything is adjusted to try to achieve the same level of contrast and remain accessible

Screenshot from 2019-07-09 08 02 52@2x
Screenshot from 2019-07-09 08 03 24@2x

D0J0P:
Maybe someday when there's a "tablet mode" for tablets and 2-in-1s that has a removable keyboard, you can add the ability to remove all keycap shortcuts to clean it up and show only the available affordance.

cassidyjames:
@D0J0P I was thinking about that! We should theoretically be able to check for attached input devices and only show them when there's a keyboard.

PR #529, opened 2019-07-25
ModelButton: only pad top level children

danrabbit:
Fixes issues with grids of labels inside of modelbuttons


cassidyjames:
@danrabbit example of where this could be seen?

danrabbit:
@cassidyjames I linked https://github.com/elementary/friends/pull/6 which is where I first noticed the issue

cassidyjames:
@danrabbit ah, missed that link from this side. Thanks, I'll take a look.

PR #533, opened 2019-08-17
Remove GTK2 theme

### LINKED ISSUE #458: Drop GTK2 theme
Since we don't support GTK2 in any way, it's not HiDPI-aware, and the theme has become massively out of sync with the GTK3 stylesheet, I wonder if we should drop it? It would also make it more obvious when apps are attempting to use GTK2 styles/variables instead of GTK3.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/67619041-drop-gtk2-theme?utm_campaign=plugin&utm_content=tracker%2F45189256&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F45189256&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cassidyjames:
Fixes #458 

~This might be terrible; it assumes "Adwaita" exists in the same directory as the elementary stylesheet. But it works: No more GTK2 theme, and it falls back to the Adwaita GTK2 theme.~


danrabbit:
I think I'm gonna close this because the plan is to have a green field branch and we just won't include a gtk2 theme there

PR #539, opened 2019-08-26
Use the same popover shadow everywhere

danirabbit:
Uses the smaller popover shadow from the panel everywhere and reduce margins (which I'm not sure actually does anything anymore, tbh)


cassidyjames:
@danrabbit what's the reason for the preference of the smaller shadow versus the larger one? This looks fine, just trying to understand the reasoning/intent.

PR #544, opened 2019-09-27
Change window shadows to be softer

lainsce:
With this PR:
- Inactive windows don't clash with active windows.
- Reduces the current visual ugliness of overlapping many windows shadows.

Looks like this:
Screenshot from 2019-09-27 01 50 28

NOTE: This doesn't fix the above for legacy windows.


danrabbit:
I'm going to close this because overlapping a bunch of exactly same size windows on top of each other is kind of a weird edge case

PR #557, opened 2019-10-23
Only apply border to SSD decoration

danrabbit:
Doesn't really fix the problem, but it is slightly better

## BEFORE
Screenshot from 2019-10-23 08 34 42@2x

## AFTER
Screenshot from 2019-10-23 08 35 20@2x

### DIFF
--- a/elementary/gtk-3.0/gtk-widgets.css
+++ b/elementary/gtk-3.0/gtk-widgets.css
@@ -4095,10 +4095,7 @@ decoration:backdrop {
 
 .ssd decoration,
 .ssd decoration:backdrop {
-    box-shadow:
-        0 0 0 1px @decoration_border_color,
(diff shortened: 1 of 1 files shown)


micahilbery:
To me it looks like the after screenshot is about half as dark.

danrabbit:
@Exalm zoom in an look at the top especially. It's a very subtle change, but it is lighter

Exalm:
Ah, indeed, sorry. Still way too dark though, it still looks like double shadow to me. Maybe there's yet another place, because here's what the default shadow is supposed to look like:

Screenshot from 2019-10-23 20-57-12

Exalm:
Huh, nevermind.

Screenshot from 2019-10-23 21-00-38

So that is the default shadow then, so the next step would be to tweak that to match the csd one.

PR #559, opened 2019-10-27
Dark Theme: Entry with more contrast

lainsce:
Currently, the dark theme entries blend with the background of everything, so they are easy to not be noticed. This solves that by making them darker.

WIP.

Current:
Screenshot from 2019-10-27 03 53 31

New:
Screenshot from 2019-10-27 03 59 06


cassidyjames:
I'm a fan of the higher contrast. I think the bigger question (i.e. for @danrabbit) is if we wanna go lighter or darker than the background for entries. Darker might make sense since they're recessed and consequently in shadow, but in the light style we have entries generally lighter than the background.

lainsce:
In my opinion, the light theme plays by different rules than the dark theme. Entries there should be white, but not only recessed/darker colored otherwise it doesn't stand out with the background...

danrabbit:
Gonna close since we're focusing on Greenfield for style changes

PR #560, opened 2019-10-29
Dark style: Matching Plank theme

lainsce:
Adds the Dark Theme also for Plank. Uses colours derived from the Black swatches of the elementary palette.

Looks like this:
Screenshot from 2019-10-29 15 35 05


cassidyjames:
:+1: on the design, but unsure if that's where it should live.

cassidyjames:
@worldofpeace right, but it's not clear how theme switching with Plank is going to work, and putting it in this folder does not make it auto-apply. It does make it appear in the (hidden) Plank settings, though, which is interesting.

cassidyjames:
Ah, I see now that it was also added to the build system to install into the Plank themes directory. In that case yeah, I think having it alongside the other dock theme in the codebase would make sense, maybe named `dock-dark.theme`. It can still be installed as `elementary-dark` if we want to do that for now.

worldofpeace:
> @worldofpeace right, but it's not clear how theme switching with Plank is going to work, and putting it in this folder does not make it auto-apply. It does make it appear in the (hidden) Plank settings, though, which is interesting.

Theme switching for plank would be using [`Plank.DockPreferences` `theme`](https://github.com/ricotz/plank/blob/dc7c2da4280a51a5cde7a803fd6ef85fb2c4618d/lib/DockPreferences.vala#L60)?
Not sure if you want that to be configurable in
* https://github.com/elementary/switchboard-plug-pantheon-shell/blob/5a66f2c502ae84157ff6bf0695bacb1fa4c4b4e5/src/Views/Dock.vala

outside the dark preference.

cassidyjames:
@worldofpeace I understand how Plank works, but I don't think we want to introduce a separate option for the dock theme specifically; we want to sync it with a system-wide dark style preference. Of course we could manually switch the theme using the Plank setting itself, but it is probably more correct for Plank itself to support the dark style preference and to support a dark variant of the theme for this purpose. Especially since the light style is included in the GTK styleseheet directory, and Plank added support for pulling the dock theme from the GTK styleseheet directory for us. :)

PR #561, opened 2019-10-30
Dark Theme: Normal buttons with more contrast

lainsce:
By making them look raised, the buttons now look more contrasted with the general background.

Looks like:
Screenshot from 2019-10-29 23 06 01

BONUS: Fixes buttons in Firefox when using Dark theme because they would get lost as the text would be white but no background.


danrabbit:
I'm gonna close this since I think we want to focus on Greenfield for style changes from now on

PR #562, opened 2019-11-01
Release 5.3.0

cassidyjames:
Bumped minor number since we've removed styles.


decathorpe:
Yeah this PR also creates an invalid appdata file with lots of stray U+200B (zero-width space) unicode characters in it ... where do you get this stuff? :grin:

cassidyjames:
@decathorpe I have no idea how that gets in there. :cry: Thanks, will fix.

PR #568, opened 2019-12-14
Tighten kerning on h1 and h2 labels

danrabbit:
Following a rabbit hole after seeing a @steveschoger tweet. The reasoning here is that since title text is larger the space between letters is also larger and makes it feel like letter spacing in titles is looser than in body text. Reducing the letter spacing in titles makes it feel like title and body text are spaced the same. Kind of interesting. Anyways, here's what it looks like in practice:

## BEFORE
Screenshot from 2019-12-13 15 59 44@2x
Screenshot from 2019-12-13 16 01 06@2x

## AFTER
Screenshot from 2019-12-13 15 59 30@2x
Screenshot from 2019-12-13 16 00 56@2x

### DIFF
--- a/elementary/gtk-3.0/granite-widgets.css
+++ b/elementary/gtk-3.0/granite-widgets.css
@@ -586,11 +586,13 @@ window.rounded decoration {
 
 .h1 {
     font-size: 24pt;
+    letter-spacing: -0.05em;
 }
 
 .h2 {
     font-weight: 300;
     font-size: 18pt;
+    letter-spacing: -0.05em;
 }
 
 .h3 {
--- a/elementary/gtk-3.0/granite-widgets.css
+++ b/elementary/gtk-3.0/granite-widgets.css
@@ -586,7 +586,7 @@ window.rounded decoration {
 
 .h1 {
     font-size: 24pt;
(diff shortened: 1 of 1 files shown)


cassidyjames:
@katiemkblaede typographical suggestions here? This code is pulling the tracking in by 5% of an em.

lainsce:
Maybe just 3% of an em instead?

danrabbit:
@cassidyjames yeah something that's kind of interesting it that it seems that these values are being rounded to the nearest pixel so for h2 it won't do anything until 5% at the default size and h1 at 4%. Maybe that's a bug upstream in Gtk that it should round to the nearest half pixel for HiDPI

cassidyjames:
Talked to @katiemkblaede in person, and she did not have a strong preference either way. She did think the `.h2` style should be somewhere inbetween the two once we can do subpixel letter-spacing—but for now the wider text is more readable so we should probably default towards that.

For `.h1` because of the bigger size and thicker weight, the tighter tracking does seem to work well. So maybe we start there, and make a note for `.h2` to come back after we have GTK 3.24+?

danrabbit:
Closing because Greenfield

PR #572, opened 2020-01-02
Improve contrast in text selections

danrabbit:
Contrast here goes from 2.8:1 to 8.8:1 in the light style and from  2.8:1 to 5.1:1 in the dark style. So oddly the effect looks less jarring but is way more contrasty

## BEFORE
Screenshot from 2020-01-02 10 15 20@2x

Screenshot from 2020-01-02 10 33 03@2x

## AFTER
Screenshot from 2020-01-02 10 13 57@2x

Screenshot from 2020-01-02 10 25 23@2x

### DIFF
--- a/data/stylesheet.appdata.xml.in
+++ b/data/stylesheet.appdata.xml.in
@@ -14,6 +14,11 @@
     <id>io.elementary.stylesheet</id>
   </provides>
   <releases>
+    <release version="5.4.1" date="2020-01-02" urgency="medium">
+      <ul>
+        <li>Improve contrast in text selections</li>
+      </ul>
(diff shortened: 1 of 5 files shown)


cassidyjames:
Should we effectively deprecate `colorAccent` in favor of `accent_color_500`? And point the former to the latter to ensure they stay in sync?

danrabbit:
@cassidyjames yeah I think that makes sense

danrabbit:
@cassidyjames I think we should address that in a different PR because it would require reworking a lot of stuff right now. That would significantly change colors in things and I think some of the gradients would be not so nice

PR #580, opened 2020-01-22
Tweaks for Inter

danrabbit:


danrabbit:
Closing because we'll use Inter with Greenfield

PR #581, opened 2020-01-22
Fix dialog label valign

danrabbit:
## Before
Screenshot from 2020-01-22 10 47 41@2x

## After
Screenshot from 2020-01-22 10 47 07@2x

### DIFF
--- a/elementary/gtk-3.0/gtk-widgets.css
+++ b/elementary/gtk-3.0/gtk-widgets.css
@@ -4209,6 +4209,10 @@ dialog toolbar {
     background-color: @bg_color;
 }
 
+.dialog-action-area button > * {
+    min-height: calc(1em + 7px);
+}
+
 /***************
 * Emoji Picker *
 ***************/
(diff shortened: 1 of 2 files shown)


elementary/gtk-3.0/gtk-widgets.css
```diff
@@ -4209,6 +4209,10 @@ dialog toolbar {
     background-color: @bg_color;
 }
 
+.dialog-action-area button > * {
+    min-height: calc(1em + 7px);
```
cassidyjames: Do we wanna comment how this was derived?

danrabbit: tbh I don't really understand why this works. I guess it's like theoretically the alignment inside the button or grid or whatever it is should be the height of the text plus padding, but then for some reason it's off by one? So maybe there's some kind of rounding? I dunno, it's magic numbers

cassidyjames: Good enough for me 😜

micahilbery: Could it be off by 1px because it's accounting for the text-shadow?

Issue #532, opened 2019-08-13
Rewrite in Sass


danrabbit:
Some of these selectors are really hard to keep track of in regular CSS

kgrubb:
here's some reasons on why we'd want to move to the sass-based stylesheet paradigm:

* A Sass build system eases creation of theme variants

* Possibly increases performance due to precompiled variables

* Shrinks filesize via compressing/minifying the gtk.css file(s)

* Makes public & private variables more straightforward, better defining the stylesheet's api

* Simplifies dealing with widget trees

danrabbit:
I'm gonna close this since it's happening and I'm not sure what more there is to track here

PR #586, opened 2020-01-31
start building via sass

kgrubb:
There's still a lot of work needed to be done, but this is just for playing around with in the meantime.

You can play around with the theme via storybook:
1. `npm i` - install dependencies
2. `npm run build` - build the css files from the sass files in the src dir.
3. `npm run storybook`- starts a webserver with storybook. You should be able to view the webpage via `localhost:6006`


btkostner:
What advantages are there to using scss right now? I see a bunch of variable changes, but gtk already has variable support, so what's the big benefit of adding another type of variable?

Also, I love storybook, and I would highly recommend it for websites, but using it to demo gtk stylesheets is going to be hard. The GTK "DOM" is pretty different from an html DOM.

lainsce:
Not him, but I think what was agreed upon of using `.scss` was Stylesheets Accents easier building, and svg hotchanging? I'd need further proof of those myself, but that's what I've read.

kgrubb:
> What advantages are there to using scss right now? I see a  bunch of variable changes, but gtk already has variable support, so  what's the big benefit of adding another type of variable?

From https://github.com/elementary/stylesheet/issues/532, It looks like there's is desire to use scss to help generate the stylesheets. If there's a different route we'd like to take, I'd be glad to shift towards whatever that may be.

[Adwaita](https://gitlab.gnome.org/GNOME/gtk/tree/master/gtk/theme/Adwaita) currently uses a similar process using `scssc` to generate the default gtk theme ([render script](https://gitlab.gnome.org/GNOME/gtk/blob/master/gtk/theme/Adwaita/render-assets.sh), [build config](https://gitlab.gnome.org/GNOME/gtk/blob/master/gtk/theme/Adwaita/meson.build)).

> Also, I love storybook, and I would highly recommend it for websites, but using it to demo gtk stylesheets is going to be hard. The GTK "DOM" is pretty different from an html DOM.

Agreed, I know they're quite different, but I'm unaware of any tooling that helps visualize gtk styling for ci purposes and for showcasing styles. If there are any projects out there, I'd love to take a look at them and perhaps use something that is geared for gtk dev instead of storybook.

btkostner:
A random idea for CI, we could use github actions to run the `gtk3-widget-factory` in CI, take screenshots, and attach them to the PR. It would probably be hard to do, but most accurate.

cassidyjames:
My feeling is that it is probably not worthwhile converting the existing stylesheet to scss, but if/when we tackle a greenfield version, we'd like to do it scss from the start. Advantages are mostly nesting, plus being able to better differentiate between internal compile-time variables (sass) and external API (GTK variables).

kgrubb:
closing in favor of the work going on in the [greenfield branch](https://github.com/elementary/stylesheet/tree/greenfield).

PR #610, opened 2020-04-24
Add window shadows

danrabbit:
This adds a neato little window shadow mixin helper and uses it for `window` elements


danrabbit:
Renamed the mixin to just "shadow" and made the selectors more specific so that we're sure we're targeting top level windows that aren't popups (aka menus)

PR #611, opened 2020-04-25
Add dialog decorations and adjust shadow mixin

danrabbit:
* Added a level3 shadow for dialogs and moved the window shadow to level4
* Redo of level2 to match the shadow we use currently in popovers because it's way nicer
* Rounded corners on windows and dialogs

### DIFF
--- a/src/widgets/_widgets.scss
+++ b/src/widgets/_widgets.scss
@@ -44,8 +44,6 @@ window:not(.popup) {
 }
 
 dialog {
-    @include shadow(-1);
-
     > decoration {
         @include shadow(3);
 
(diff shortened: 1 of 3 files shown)


src/widgets/_widgets.scss
```diff
@@ -53,6 +53,11 @@ dialog {
             @include shadow(2);
         }
     }
+
+    > headerbar.titlebar label.title {
+        font-size: 0.1px;
```
cassidyjames, line 56: Does this still not work correctly in more recent versions of GTK if we just set 0px?

danrabbit: I'm not sure, but I can't remember exactly what the issue was before either. I can add a comment

PR #612, opened 2020-04-26
Dry imports

danrabbit:
It's gonna be too much maintenance to update these color scheme files each time we add a new import, so instead:

* move mixins to /src because they're not really widgets right?
* Have mixins import widgets

### DIFF
--- a/src/_mixins-dark.scss
+++ b/src/_mixins-dark.scss
@@ -23,3 +23,5 @@
             0 10px 10px rgba(0, 0, 0, 0.32);
     }
 }
+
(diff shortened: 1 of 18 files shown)


cassidyjames:
So yeah, I have been wondering about the naming and structure of things in general here. I wonder if it makes sense to just have like `main.scss` instead of `mixins.scss`, especially since there's only one mixin there? I'm not sure. I guess I can merge this (since it's an improvement) and then we can ask someone who probably has more experience and thoughts with Sass like @btkostner?

PR #614, opened 2020-04-28
Add Spinner styles

danrabbit:


cassidyjames:
@danrabbit so I'm not seeing the spinner spin at all in AppCenter with this branch, but I am wondering if it's a difference between GTK versions in bionic vs. focal.

danrabbit:
Hm yeah it's working as expected here

PR #620, opened 2020-04-29
Style window titlebars

danrabbit:
I switched the shadow mixin to a function so that we can append other shadows to it (like highlights or border shadow).

This should resolve missing border shadow and make things a bit nicer overall


src/_mixins.scss
```diff
@@ -66,4 +91,57 @@ $toplevel-border-color: rgba(0, 0, 0, 0.3);
...
+        );
+    box-shadow:
+        inset 0 1px 0 0 #{'alpha(@highlight_color, 0.3)'},
+        inset 1px 0 0 0 #{'alpha(@highlight_color, 0.07)'},
+        inset -1px 0 0 0 #{'alpha(@highlight_color, 0.07)'},
+        inset 0 -1px 0 0 #{'alpha(@highlight_color, 0.2)'},
+        0 0 0 1px rgba(0, 0, 0, 0.15),
+        0 1px 2px rgba(0, 0, 0, 0.2);
+
+    &.titlebar {
```
danrabbit, line 110: I'm not sure if it makes sense to keep this here so that we always sync these values or if this should go in widgets since it's specific to that one widget

cassidyjames: I _think_ it would make more sense in widgets since it's specific to a widget and not generally useful.

src/_mixins.scss
```diff
@@ -1,17 +1,42 @@
...
+$bg-color-2: $SILVER_100;
+
+// Sidebars and inline toolbars
+$bg-color-3: mix($SILVER_100, $SILVER_300, $weight: 15%);
+
+// Titlebars and toolbars
+$bg-color-4: mix($SILVER_100, $SILVER_300, $weight: 30%);
+
+// Text, images, and other foreground elements
+$fg-color: $BLACK_500;
```
cassidyjames, line 17: I wonder if it makes more sense to do these like the shadow levels as a mixin instead of individual variables? I.e. you're setting a bg color with elevation 1, 2, 3, etc. Otherwise, I think I'd want to look at more semantic names for the variables.

src/widgets/_widgets.scss
```diff
@@ -29,38 +29,64 @@ dialog {
...
     }
 
     &.tiled {
         > decoration {
-            @include border-shadow;
             @include shadow(1);
         }
     }
+
+    headerbar.titlebar {
```
cassidyjames, line 48: Is there a reason this is specific to headerbar.titlebar and no just .titlebar?

cassidyjames: I'm also not sure that it makes sense to nest this here, since `.popup` isn't going to have a headerbar widget… I think it would make more sense for this to be top-level.

danrabbit: `.titlebar` only is fine, but these styles are specifically for titlebars on Windows and don't apply to dialogs

PR #622, opened 2020-04-29
Add label styles

danrabbit:


danrabbit:
Hm that’s an interesting point. I guess it doesn’t need to apply only to labels

PR #625, opened 2020-05-02
Add stacksidebar style

danrabbit:
Includes a lot of the same changes to mixins as #620 

Styles the basic sidebar as seen in Granite Demo. Not intended to do all the styling required for Files (for example)

### DIFF
--- a/src/_mixins.scss
+++ b/src/_mixins.scss
@@ -1,18 +1,55 @@
+// Inset box shadows
+@define-color highlight_color #fff;
+
+// The default titlebar color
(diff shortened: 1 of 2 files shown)


cassidyjames:
Is this sort of sidebar used anywhere else? I'm struggling to find where else it would show up. :sweat_smile: But it looks fine as a start to me. :shrug:

danrabbit:
I’m not sure we use it in any of our apps, but it should set up basic sidebar styles that can be built on

PR #626, opened 2020-05-04
Add entry styles

danrabbit:
I changed the background color mixin to a function so that we could use the values with `mix`

I also tweaked the dark titlebar colors a little bit since it felt really dark after adding the entry


cassidyjames:
@danrabbit do you want to do focus styles in this branch, or in a follow-up?

danrabbit:
@cassidyjames I’d like to revisit focus styles separately since I think there’s some more design work needed there.  My goal right now is just to get the stylesheet to a minimally usable state

PR #629, opened 2020-05-05
Greenfield: separate into component files

cassidyjames:
Didn't touch color variants yet, probably will do so in a follow-up.


src/_mixins.scss
```diff
@@ -155,4 +155,5 @@ $toplevel-border-color: rgba(0, 0, 0, 0.2);
 }
 
 @import '_animate.scss';
-@import 'widgets/_widgets.scss';
+@import '_typography.scss';
+@import 'widgets/_index.scss';
```
btkostner, line 159: Does this work?

```suggestion
@import 'widgets';
```

PR #630, opened 2020-05-05
Greenfield: Use ems for padding

cassidyjames:
Adds an `em()` function that takes a size and converts it to ems. The second argument is a font-size if you have a different context and know you want a different font size to calculate from, otherwise it defaults to elementary defaults which is 9pt.

If you don't pass in a unit, it assumes `px`.

### DIFF
--- a/src/_mixins.scss
+++ b/src/_mixins.scss
@@ -135,6 +135,18 @@ $toplevel-border-color: rgba(0, 0, 0, 0.2);
     @return $highlight;
 }
 
+@function em($pixels, $text-size: 9pt) {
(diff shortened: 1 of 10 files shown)


danrabbit:
This should probably be for corner radius too right?

cassidyjames:
I don't know! Definitely not for strokes, but I was unsure about corner radii in general.

danrabbit:
I would think yes for radius if the goal is to keep things looking relatively the same just bigger/smaller

Looks like you missed titlebar padding in windows also.

PR #645, opened 2020-05-07
em(): Use rems to avoid errors with small text

danrabbit:


cassidyjames:
Probably should rename to `rem()` for clarity while we're at it?

PR #647, opened 2020-05-07
Greenfield: Simplify variant directory structure, install Plank themes

### LINKED ISSUE #641: Add plank themes
The new greenfield stylesheets don't include the plank theme

cassidyjames:
Fixes #641. Since https://github.com/mesonbuild/meson/issues/2320 does not allow subdirs when building, @tintou did some Python magic. :heart: 

Note that **+128** lines of this diff are just the Plank files.


src/meson.build
```diff
@@ -1,8 +1,51 @@
...
+        build_by_default: true
+    )
+    variant_dark_target = custom_target(
+        meson.project_name() + '.' + variant + '-dark',
+        input : 'variants' / variant + '-dark.scss',
+        output : variant + '-dark.css',
+        command : sass_command,
+        build_by_default: true
+    )
+    install_exec = meson.current_source_dir() / '..' / 'meson' / 'install-to-dir.py'
```
danrabbit, line 32: How about
```suggestion
    install_exec = join_paths(meson.source_root (), 'meson', 'install-to-dir.py')
```

PR #653, opened 2020-05-07
Add styles for color buttons

danrabbit:
Screenshot from 2020-05-07 16 07 00@2x
Screenshot from 2020-05-07 16 07 09@2x


cassidyjames:
Since every one of these starts with `.color-button`, it might make sense to just nest it all under there?

PR #657, opened 2020-05-08
Tooltips and granite.overlaybar

danrabbit:


src/widgets/_tooltips.scss
```diff
@@ -0,0 +1,20 @@
+tooltip.background,
+.overlay-bar {
+    background-color: #{'alpha (#222, 0.9)'};
```
cassidyjames, line 3: Any way to avoid hardcoding a hex value here?

src/widgets/_tooltips.scss
```diff
@@ -0,0 +1,20 @@
...
+    border-radius: rem(3px);
+    padding: rem(3px) rem(6px);
+
+    label,
+    spinner {
+        color: white;
+        text-shadow: 0 rem(1px) rem(2px) rgba(0, 0, 0, 0.6);
+        -gtk-icon-shadow: 0 rem(1px) rem(2px) rgba(0, 0, 0, 0.6);
```
cassidyjames, line 11: I think I want to start doing this for clarity that we're intentionally picking "literal black" instead of what could look like some arbitrary color at a glance.

```suggestion
        text-shadow: 0 rem(1px) rem(2px) rgba(black, 0.6);
        -gtk-icon-shadow: 0 rem(1px) rem(2px) rgba(black, 0.6);
```

src/widgets/_tooltips.scss
cassidyjames: At the very least, you can do this so it's compiled; it doesn't need to be runtime alpha.

```suggestion
    background-color:  rgba(#222, 0.9);
```

PR #658, opened 2020-05-08
Add linked styles

danrabbit:


src/widgets/_linked.scss
```diff
@@ -0,0 +1,98 @@
...
+    &:not(:first-child) {
+        @extend %linked_not_top;
+    }
+
+    &:not(:last-child) {
+        @extend %linked_not_bottom;
+    }
+}
+
+combobox.linked button:nth-child(2) {
```
cassidyjames, line 57: Is there some amazing GTK reason this isn't `:last-child`?

danrabbit: Yeah Comboboxes are fuckin weird. I wouldn't mess with it

danrabbit:
@cassidyjames I guess Gtk must have gotten smarter/easier. Removing all the cases for RTL magically makes it work

PR #661, opened 2020-05-08
Support rounded cards

### LINKED ISSUE #640: Add .rounded support to cards
Should probably also demo in Granite.

danrabbit:
As seen on the greeter and in appearance settings

Fixes #640


danrabbit:
@cassidyjames probably not because different things are going to have different corner radii anyways even if we support multiple widgets with it

PR #663, opened 2020-05-08
Switches: Get rid of labels in GTK 3.22

cassidyjames:
So we had said we want to ignore styling for Hera, but this has been annoying me for screenshots and it’s just one line, with a comment. :shrug:


danrabbit:
I mean, I'd rather not carry legacy styles because once you start... There's gonna end up being a lot of breakage on Hera, especially because I'm uncovering lots of weird things like the notebook spinner and I submitted a branch for the switches in wingpanel. I think it's better to really start clean here and focus on Odin

PR #672, opened 2020-05-10
Experimental: Dark accent tint

cassidyjames:
Just seeing if we could tint the dark style with the accent color to bring that color subtly into more of the UI. Unsure how I feel about it, but it's kind of a neat effect. This mixes just 1% of the accent color's darkest shade into all dark bg levels, and 50% of the lightest shade into the highlight color. We could go less subtle, but I thought this was a good starting point.

To do so, I had to add sass variables for the accent colors. One drawback I see: this effect is not dynamic with application-set accent colors. The app would either need to request a (new, non-existent yet) neutral color scheme and then set its own accent, or be okay with the subtle user tint. The reason I could not use the GtkCSS variable is that we use the bg colors in other sass functions down the line.

I like the idea of an almost imperceptible tint for the dark style, but I don't know if the effect is worth the effort.

End result:

Screenshot from 2020-05-10 14 14 24@2x

Screenshot from 2020-05-10 14 14 29@2x

### DIFF
--- a/src/_mixins.scss
+++ b/src/_mixins.scss
@@ -11,7 +11,7 @@ $fg-color: $BLACK_500;
 $toplevel-border-color: rgba(0, 0, 0, 0.2);
 
 @if $color-scheme == "dark" {
-    @define-color highlight_color #{'alpha(' + mix(white, $accent_color_100, 50%) + ', 0.3)'};
+    @define-color highlight_color #{'alpha(' + mix(white, $accent_color_100) + ', 0.2)'};
 
     $titlebar-color: mix(mix($BLACK_500, $BLACK_700, 70%), $accent_color_900, $weight: 99%);
 
--- a/src/_mixins.scss
+++ b/src/_mixins.scss
@@ -11,9 +11,9 @@ $fg-color: $BLACK_500;
 $toplevel-border-color: rgba(0, 0, 0, 0.2);
 
 @if $color-scheme == "dark" {
-    @define-color highlight_color rgba(255, 255, 255, 0.2);
+    @define-color highlight_color #{'alpha(' + mix(white, $accent_color_100, 50%) + ', 0.3)'};
 
-    $titlebar-color: mix($BLACK_500, $BLACK_700, $weight: 70%);
+    $titlebar-color: mix(mix($BLACK_500, $BLACK_700, 70%), $accent_color_900, $weight: 99%);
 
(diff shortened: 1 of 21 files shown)


lainsce:
To get the recolorable tint for third-party apps, scss has the `#{};` (Interpolation) operator; using this on parts of the SCSS code that should become recolorable will do the trick. Found about this operator today. :man_scientist: 

So maybe:
```scss
(...)
@if $level == 1 {
    @return #{"mix($BLACK_300, $accent_color_900, 99%)"};
} @else if $level == 2 {
    @return #{"mix($BLACK_500, $accent_color_900, 99%)"};
} @else if $level == 3 {
    @return #{"mix(mix($BLACK_500, $BLACK_700, 75%), $accent_color_700, 99%)"};
(...)
```
this will probably give 3rd-party accents access to the tint in this PR.
(Untested.)

cassidyjames:
@lainsce as mentioned in the PR, I had to change this to compile-time because we use the output here later on with sass color functions. The only way around that would be to change everything to GtkCSS color functions, losing all of the benefits of sass along the way.

lainsce:
I only see some places where this is used inside color functions,
- L15 on `_index.scss`
- L4 on `_entries.scss`

Are those the only places?

lainsce:
I've tested something and it works, maybe will help (app developers would just need to override those three tint level GTKCSS named colors):

```scss
// Tints
$tint-level-one: mix($BLACK_300, $accent_color_900, 99%);
$tint-level-two: mix($BLACK_500, $accent_color_900, 99%);
$tint-level-three: mix(mix($BLACK_500, $BLACK_700, 75%), $accent_color_700, 99%);
@define-color color_tint_level_one #{$tint-level-one};
@define-color color_tint_level_two #{$tint-level-two};
@define-color color_tint_level_three #{$tint-level-three};

@function bg-color($level) {
(...)
    } @else if $color-scheme == "dark" {
        @if $level == 1 {
            @return $tint_level_one;
        } @else if $level == 2 {
            @return $tint-level-two;
        } @else if $level == 3 {
            @return $tint-level-three;
        } @else if $level == 4 {
            @return $titlebar-color;
        }
    }
}
```

cassidyjames:
This was a fun experiment, but I don't see it being worthwhile for how much extra work it makes. We can revisit it fresh later if we determine we want to tint apps with the accent color.

PR #677, opened 2020-05-11
Animate: Use transition times and easing from Material

danrabbit:
Instead of making up values

### DIFF
--- a/src/_animate.scss
+++ b/src/_animate.scss
@@ -1,3 +1,23 @@
+@function duration($transition) {
+    @if $transition == "open" {
+        @return 250ms;
+    } @else if $transition == "close" {
+        @return 200ms;
+    }
+
+    @if $transition == "expand" {
+        @return 300ms;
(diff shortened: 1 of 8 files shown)


src/_animate.scss
```diff
@@ -1,3 +1,23 @@
...
+    } @else if $transition == "collapse" {
+        @return 250ms;
+    }
+
+    @if $transition == "element" {
+        @return 100ms;
+    }
+}
+
+$easing: cubic-bezier(0.4, 0, 0.2, 1);
```
cassidyjames, line 19: GTK doesn't seem to love this… it seems like it is less smooth than the `ease-out` in the scrollbar, especially—I see it animating the last frame much slower than the previous easing. Not sure if it matters that much, but it's gonna bother me. :stuck_out_tongue:

danrabbit:
I'd like to try it and see and we can always come back and iterate

PR #680, opened 2020-05-11
Add checkerboard background class

danrabbit:
Not necessarily always used with cards, but mostly. Not sure if there's a better place for this


cassidyjames:
I say we either add a checkerboards file, or a style_classes one for generic/reusable classes? Or maybe a granite file?

cassidyjames:
Or honestly just in index for now until we have a better home 🤷

PR #685, opened 2020-05-12
Titlebars: add etched labels and icons

cassidyjames:
This is a stylistic thing I really love have been missing: the subtle etched look on window titles. It's also being applied to icons which looks great on window controls and symbolic icons i.e. in Terminal, but I'm not sure if it needs to be made more specific to not apply to hicolor icons, or child widgets (i.e. buttons in popovers).

Screenshot from 2020-05-12 22 59 15@2x
Screenshot from 2020-05-12 22 59 20@2x

It also might be too harsh/need to be made more subtle.

~If this is a stylistic thing we pull in, I figure eventually we might want to pull it into a placeholder class to keep it consistent—but since it's only used in one place here, I didn't want to over-abstract it.~

### DIFF
--- a/src/_mixins.scss
+++ b/src/_mixins.scss
@@ -144,7 +144,7 @@ $toplevel-border-color: rgba(black, 0.2);
 
 @function etched-shadow() {
     @if $color-scheme == "dark" {
-        @return 0 -1px 1px rgba(black, 0.67);
+        @return 0 -1px 1px #{'shade (@color_primary, 0.75)'};
     }
 
     @return 0 1px 1px #{'@highlight_color'};
--- a/src/_mixins.scss
+++ b/src/_mixins.scss
@@ -142,6 +142,14 @@ $toplevel-border-color: rgba(black, 0.2);
     }
 }
 
+@function etched-shadow() {
+    @if $color-scheme == "dark" {
+        @return 0 -1px 1px rgba(black, 0.67);
+    }
+
+    @return 0 1px 1px #{'@highlight_color'};
+}
+
 @function rem($pixels, $text-size: 9pt) {
     @if (unitless($pixels)) {
         $pixels: $pixels * 1px;
--- a/src/_mixins.scss
+++ b/src/_mixins.scss
@@ -142,14 +142,6 @@ $toplevel-border-color: rgba(black, 0.2);
     }
 }
 
-@function etched-shadow() {
-    @if $color-scheme == "dark" {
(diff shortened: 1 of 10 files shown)


cassidyjames:
Okay, tweaked the specificity. Direct children is unreliable because if you use a grid or other container, you lose the styles. So I styled specifically the title (we don't typically have text in titlebars other than the title anyway), and then explicitly titlebuttons, image-buttons, and mode-swiches (since they use icons). This avoids accidentally etching anything inside a button that shouldn't be. 

I also explicitly reset in child popovers and menus so we don't traverse farther than the titlebar widget.

src/_mixins.scss
```diff
@@ -142,6 +142,14 @@ $toplevel-border-color: rgba(black, 0.2);
     }
 }
 
+@function etched-shadow() {
```
danrabbit, line 145: Kind of seems like premature optimization to put this here when it's only used in `_windows.scss`

src/_mixins.scss
```diff
@@ -142,6 +142,14 @@ $toplevel-border-color: rgba(black, 0.2);
...
+    @if $color-scheme == "dark" {
+        @return 0 -1px 1px rgba(black, 0.67);
```
danrabbit, line 147: This probably needs to be a shade of the primary color so that you don't get muddy colors with greens and yellows. Something like
```suggestion
        @return 0 -1px 1px #{'shade (@color_primary, 0.75)'};
```

Which also really only makes sense in titlebars

src/_mixins.scss
cassidyjames: Yeah I went back and forth. Since I was using it in multiple places in that file (for text and icons), I pulled it into mixins, but I can just define this function in the windows file itself.

src/widgets/_windows.scss
```diff
@@ -90,11 +90,33 @@ window:not(.popup):not(.menu) {
     }
 
     .titlebar {
+        @function etched-shadow() {
+            @if $color-scheme == "dark" {
+                @return 0 -1px 1px #{'shade (@color_primary, 0.75)'};
```
danrabbit, line 95: pwease :)

```suggestion
                @return 0 1px 1px #{'shade (@color_primary, 0.75)'};
```

cassidyjames:
Alright, simplified this a bit. It's using alpha on white/black. I don't feel strongly whether we use this or `shade` if you think one looks better. I'm kinda stuck in this in-between where the current light style looks imperceptibly subtle with the default titlebar color, but is on the heavier side with brighter brand colors.

PR #688, opened 2020-05-12
Add accent label and image support

danrabbit:


src/_typography.scss
```diff
@@ -31,6 +31,14 @@ label.h4,
     font-size: 1.22em
 }
 
+.accent {
+    color: #{'@accent_color_700'};
+
+    @if $color-scheme == "dark" {
+        color: #{'mix(@accent_color_300, @accent_color_500, 0.5)'};
```
cassidyjames, line 38: What's the reasoning for not just using `@accent_color_300` here? Did it feel too bright? I feel like this is too dark/low-contrast with cocoa and slate, but that might mean we should revisit those palettes.

danrabbit: Yeah 300 was really bright. The mix gives me about the same AA contrast for blueberry. I agree cocoa and slate are dark.

Another path would could take is setting an accent variable and do something like `@if $accent == "cocoa" {}`

danrabbit: We have a similar problem with banana in the light style which doesn't even pass A level contrast

cassidyjames: @danrabbit up to you on which is easier. Special-casing the "darker" palettes might be easier than reworking them altogether, though, at least for now. :)

danrabbit:
All colors should pass at least A level, with most passing AA

src/_typography.scss
```diff
@@ -33,8 +33,25 @@ label.h4,
 .accent {
     color: #{'@accent_color_700'};
 
+    @if $accent-name == "cocoa" or
```
cassidyjames, line 32:
```suggestion
    // Special-case especially light or dark palette colors
    @if $accent-name == "cocoa" or
```

PR #695, opened 2020-05-14
Window: square bottom corners

danrabbit:
We don't have window clipping in gtk3, so don't try to round window corners by default


src/widgets/_windows.scss
```diff
@@ -13,6 +10,10 @@ dialog {
 }
 
 window:not(.popup):not(.menu) {
+    decoration {
+        border-radius: rem(6px) 0 0;
```
cassidyjames, line 14: Did you mean

```suggestion
        border-radius: rem(6px) rem(6px) 0 0;
```

PR #706, opened 2020-05-20
Add support for Hdy.Window

danrabbit:
Can be tested with https://github.com/elementary/switchboard/pull/161

### DIFF
--- a/src/widgets/_windows.scss
+++ b/src/widgets/_windows.scss
@@ -91,7 +91,7 @@ window:not(.popup):not(.menu) {
     }
 
     &.unified {
-        decoration
+        decoration,
         decoration-overlay {
             border-radius: rem(6px);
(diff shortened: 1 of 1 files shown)


Exalm:
2 possible solutions for draggable headerbar:

1. Use `HdyHeaderBar` instead
2. Wrap it into a `HdyWindowHandle`

src/widgets/_windows.scss
```diff
@@ -90,6 +90,32 @@ window:not(.popup):not(.menu) {
         }
     }
 
+    &.unified {
```
cassidyjames, line 93: Probably good to comment this since I wouldn't know where these classes were coming from otherwise

```suggestion
    // Hdy.Window
    &.unified {
```

PR #713, opened 2020-05-25
Add Terminal class

lainsce:
As seen on faux-terminals thru the UI.

EDIT: was just in the wrong place

### DIFF
--- a/src/widgets/_views.scss
+++ b/src/widgets/_views.scss
@@ -17,4 +17,18 @@ scrolledwindow {
 .view {
     background-color: bg-color(1);
     color: $fg-color;
+
+    &.terminal {
+        background-color: #252e32;
+        color: #94a3a5;
+
+        & text {
+            font-family: monospace;
+
+            & selection {
+                background-color: #93a1a1;
+                color: #252e32;
+            }
+        }
+    }
 }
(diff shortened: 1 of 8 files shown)


src/widgets/_views.scss
```diff
@@ -17,4 +17,18 @@ scrolledwindow {
 .view {
     background-color: bg-color(1);
     color: $fg-color;
+
+    &.terminal {
+        background-color: #252e32;
+        color: #94a3a5;
+
+        & text {
+            font-family: monospace;
```
danrabbit, line 26: Looks like this actually needs to apply to `.terminal` , not `text` to take effect

lainsce: Fixed.

src/widgets/_views.scss
```diff
@@ -17,4 +17,17 @@ scrolledwindow {
...
     background-color: bg-color(1);
     color: $fg-color;
+
+    &.terminal {
+        background-color: #252e32;
+        color: #94a3a5;
+        font-family: monospace;
+
+        text {
+            selection {
```
danrabbit, line 27: can probably either remove `text` or combine this to be `text selection {}`

lainsce: Combined they work. Fixed.

PR #714, opened 2020-05-25
Add third-party expected GTK color names

lainsce:
These are seen in other apps than elementary's; adapted some to be different for dark style too.

### DIFF
--- a/src/widgets/_index.scss
+++ b/src/widgets/_index.scss
@@ -26,6 +26,7 @@ messagedialog,
 @define-color wm_shadow #{rgba(black, 0.2)};
 
 // Things other apps expect
+@define-color base_color #{"" + bg_color(1)};
 @define-color theme_text_color #{'@text_color'};
 @define-color theme_bg_color #{'@bg_color'};
 @define-color theme_base_color #{'@base_color'};
@@ -43,8 +44,8 @@ messagedialog,
 //
 
 selection {
-    background-color: #{'alpha(@accent_color_100, 0.3)'};
-    color: #{'@accent_color_900'};
+    background-color: #{'@selected_bg_color'};
+    color: #{'@selected_fg_color'};
 }
 
 @if $color-scheme == "dark" {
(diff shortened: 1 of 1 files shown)


src/widgets/_index.scss
```diff
@@ -25,6 +25,23 @@ messagedialog,
 @define-color theme_selected_fg_color #{"" + $fg-color};
 @define-color wm_shadow #{rgba(black, 0.2)};
 
+// Things other apps expect
+@define-color theme_text_color #{'@text_color'};
+@define-color theme_bg_color #{'@bg_color'};
+@define-color theme_base_color #{'@base_color'};
```
danrabbit, line 31: I don't think this is defined yet

src/widgets/_index.scss
```diff
@@ -25,6 +25,23 @@ messagedialog,
...
+@define-color selected_bg_color #{'alpha(@accent_color_100, 0.3)'};
+@define-color selected_fg_color #{'@accent_color_900'};
```
danrabbit, line 33: Probably should update the `selection` element to use these values so they always match

src/widgets/_index.scss
```diff
@@ -25,6 +25,23 @@ messagedialog,
...
+@define-color theme_selected_bg_color #{'@selected_bg_color'};
+@define-color insensitive_fg_color #{'@insensitive_color'};
+
+@if $color-scheme == "dark" {
+    @define-color selected_bg_color #{'alpha(@accent_color_700, 0.3)'};
+    @define-color selected_fg_color #{'@accent_color_100'};
+    @define-color theme_selected_bg_color #{'@selected_bg_color'};
+    @define-color theme_selected_fg_color #{'@selected_fg_color'};
```
danrabbit, line 41: If these are already defined this way normally, then this isn't necessary to define twice

lainsce: It's how it was on the old stylesheet, should I just delete them?

src/widgets/_index.scss
lainsce: Will do.

src/widgets/_index.scss
lainsce: I'll define it as `bg-color(1)` to match previous styling on old stylesheet.

PR #715, opened 2020-05-25
Add Separator styling

lainsce:
Wherever they are, they had no styling completed. This PR does that.

### DIFF
--- a/src/widgets/_index.scss
+++ b/src/widgets/_index.scss
@@ -25,20 +25,21 @@ messagedialog,
 @define-color theme_selected_fg_color #{"" + $fg-color};
 @define-color wm_shadow #{rgba(black, 0.2)};
 
-selection {
-    background-color: #{'alpha(@accent_color_100, 0.3)'};
-    color: #{'@accent_color_900'};
-}
+@define-color selected_bg_color #{'alpha(@accent_color_100, 0.3)'};
+@define-color selected_fg_color #{'@accent_color_900'};
 
 @if $color-scheme == "dark" {
     @define-color error_color @STRAWBERRY_300;
     @define-color success_color @LIME_300;
     @define-color warning_color @BANANA_100;
 
-    selection {
-        background-color: #{'alpha(@accent_color_700, 0.3)'};
-        color: #{'@accent_color_100'};
-    }
+    @define-color selected_bg_color #{'alpha(@accent_color_700, 0.3)'};
+    @define-color selected_fg_color #{'@accent_color_100'};
+}
+
+selection {
+    background-color: #{'@selected_bg_color'};
+    color: #{'@selected_fg_color'};
 }
 
 .checkerboard {
@@ -69,6 +70,14 @@ selection {
     border-radius: 9999px;
 }
 
+.avatar {
+    border: 1px solid rgba(black, 0.35);
+    border-radius: 50%;
+    box-shadow:
+        outset-highlight(),
+        outset-shadow(3);
+}
+
 .error {
     color: #{'@error_color'};
 }
@@ -87,6 +96,7 @@ selection {
 @import '_entries.scss';
 @import '_levelbars.scss';
 @import '_linked.scss';
+@import '_lists.scss';
 @import '_menus.scss';
 @import '_notebooks.scss';
 @import '_notifications.scss';
--- a/src/widgets/_index.scss
+++ b/src/widgets/_index.scss
@@ -36,6 +36,8 @@ messagedialog,
     @define-color success_color @LIME_300;
     @define-color warning_color @BANANA_100;
 
+    @define-color menu_separator rgba(black, 0.25);
+
     @define-color selected_bg_color #{'alpha(@accent_color_700, 0.3)'};
     @define-color selected_fg_color #{'@accent_color_100'};
 }
--- a/src/widgets/_index.scss
+++ b/src/widgets/_index.scss
@@ -14,6 +14,9 @@ messagedialog,
 // Placeholder text in entries. Doesn't respect alpha
 @define-color placeholder_text_color #{"" + mix($fg-color, bg-color(1), $weight: 75%)};
 
+// Separator color
+@define-color menu_separator #{"" + rgba(black, 0.15)};
+
 @define-color error_color @STRAWBERRY_500;
 @define-color success_color @LIME_700;
 @define-color warning_color @BANANA_900;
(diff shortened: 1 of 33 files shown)


src/widgets/_menus.scss
```diff
@@ -5,6 +5,16 @@
 
     separator {
         min-height: rem(6px);
+
+        &.horizontal {
+            border-top: 1px solid rgba(black, 0.15);
```
danrabbit, line 10: I think this probably needs to be added to index as `menu_separator` so that we can make sure to keep it consistent

src/widgets/_menus.scss
```diff
@@ -5,6 +5,16 @@
...
+            margin-bottom: rem(-1px);
```
danrabbit, line 11: If you're not using `rem` for the border, we probably don't want to use `rem` here

src/widgets/_panes.scss
```diff
@@ -21,4 +21,39 @@ paned.horizontal > separator {
...
+    &.wide {
+        background-image:
+            linear-gradient(
+                to bottom,
+                rgba(black, 0.15) 1px,
+                rgba(white, 0.01) 1px,
+                rgba(white, 0.01) 2px,
+                transparent 2px
+            );
+    }
```
danrabbit, line 34: This is an entirely different thing and should be in its own branch

src/widgets/_panes.scss
```diff
@@ -21,4 +21,39 @@ paned.horizontal > separator {
...
```
danrabbit, line 58: Yeah same here, these pane handles should be a totally different branch

src/widgets/_toolbars.scss
```diff
@@ -2,3 +2,15 @@ actionbar,
...
     background-color: bg-color(3);
 }
+
+toolbar separator {
+    background: none;
+    margin: 0 rem(18px);
+}
+
+toolbar paned separator,
+.toolbar .pane-separator {
```
danrabbit, line 12: where is this selector used? This seems oddly specific

src/widgets/_toolbars.scss
```diff
@@ -2,3 +2,15 @@ actionbar,
 .search-bar {
     background-color: bg-color(3);
 }
+
+toolbar separator {
```
danrabbit, line 6: Do you have an example of where this is used?

src/widgets/_windows.scss
```diff
@@ -117,6 +117,23 @@ window:not(.popup):not(.menu) {
...
+            box-shadow: none;
+        }
+
+        separator {
+            background: none;
+            margin: 0 rem(18px);
+        }
+
+        separator:first-child,
+        separator:last-child {
```
danrabbit, line 132: If this is meant for the separators next to titlebuttons then a better selector would be `
```suggestion
        separator.titlebutton {
```

src/widgets/_windows.scss
```diff
@@ -117,6 +117,23 @@ window:not(.popup):not(.menu) {
...
+            margin: 0;
+            min-width: 0;
+            padding: 0;
```
danrabbit, line 135: Is there a reason to believe that separators have any of these properties set by default? So far it looks like the only rules we have added for separators have been nested inside other elements

src/widgets/_windows.scss
```diff
@@ -117,6 +117,23 @@ window:not(.popup):not(.menu) {
             text-shadow: none;
             -gtk-icon-shadow: none;
         }
+
+        pane-separator {
```
danrabbit, line 121: I don't think this is a valid selector. What is this intended to style?

src/widgets/_windows.scss
```diff
@@ -117,6 +117,23 @@ window:not(.popup):not(.menu) {
...
+            background-color: rgba(black, 0.0);
+            box-shadow: none;
+        }
+
+        separator {
+            background: none;
```
danrabbit, line 127: Is there a reason to believe that `.titlebar separator` has a background set?

src/widgets/_menus.scss
lainsce: Adding it there.

src/widgets/_menus.scss
lainsce: Oh, missed a place, wanted to use rem everywhere here. oops.

src/widgets/_panes.scss
lainsce: Will remove.

src/widgets/_panes.scss
lainsce: Will also remove.

src/widgets/_toolbars.scss
lainsce: Some apps use this to make "dividers" in their toolbars.

src/widgets/_toolbars.scss
lainsce: Some apps use this to make "dividers" in their toolbars.

src/widgets/_windows.scss
lainsce: Oh this was found in some GNOME apps.

src/widgets/_windows.scss
lainsce: Some apps use it to make dividers, and those for some reason have a grey background if left unstyled..

src/widgets/_windows.scss
lainsce: With these removed, the separator is thin like it was in the previous stylesheet.

lainsce:
Lemme give this PR a diet.

src/widgets/_index.scss
```diff
@@ -14,6 +14,9 @@ messagedialog,
 // Placeholder text in entries. Doesn't respect alpha
 @define-color placeholder_text_color #{"" + mix($fg-color, bg-color(1), $weight: 75%)};
 
+// Separator color
+@define-color menu_separator #{rgba(black, 0.15)};
```
danrabbit, line 18: You don't need to unquote this

```suggestion
@define-color menu_separator rgba(black, 0.15);
```

src/widgets/_menus.scss
```diff
@@ -5,6 +5,16 @@
 
     separator {
         min-height: rem(6px);
+
+        &.horizontal {
+            border-top: rem(1px) solid #{'@menu_separator'};
+            margin-bottom: rem(-1px);
```
danrabbit, line 11: These should probably not be in rems since this is a stroke

lainsce:
All fixed now.

PR #716, opened 2020-05-26
Make Selection visible everywhere

lainsce:
This target `*:selected` should capture all selected states and as bonus help with some broken UI states currently present with the stylesheet.

It's overridable by using the `&:selected` block on the widget that needs to be overridden.

### DIFF
--- a/src/_mixins.scss
+++ b/src/_mixins.scss
@@ -192,6 +192,16 @@ $toplevel-border-color: rgba(black, 0.2);
     }
 }
 
+%selected_item {
+    @if $color-scheme == "dark" {
+        background-color: #{"@selected_bg_color"};
+        color: #{"@selected_fg_color"};
+    } @else if $color-scheme == "light" {
+        background-color: #{"@selected_bg_color"};
+        color: #{"@selected_fg_color"};
+    }
+}
+
 @import '_animate.scss';
 @import '_typography.scss';
 // Can't just use `@import 'widgets';` because of sass version in repos
--- a/src/_mixins.scss
+++ b/src/_mixins.scss
@@ -193,13 +193,8 @@ $toplevel-border-color: rgba(black, 0.2);
 }
 
 %selected_item {
-    @if $color-scheme == "dark" {
-        background-color: #{"@selected_bg_color"};
-        color: #{"@selected_fg_color"};
-    } @else if $color-scheme == "light" {
-        background-color: #{"@selected_bg_color"};
-        color: #{"@selected_fg_color"};
-    }
+    background-color: #{"@selected_bg_color"};
+    color: #{"@selected_fg_color"};
 }
 
 @import '_animate.scss';
--- a/src/_mixins.scss
+++ b/src/_mixins.scss
@@ -17,7 +17,7 @@ $toplevel-border-color: rgba(black, 0.2);
 
     $fg-color: white;
 
-    $toplevel-border-color: rgba(black, 0.6);
+    $toplevel-border-color: rgba(black, 0.75);
 }
 
 // The primary brand color used for titlebars
--- a/src/widgets/_entries.scss
+++ b/src/widgets/_entries.scss
@@ -14,6 +14,10 @@ entry {
         color: #{'@placeholder_text_color'};
     }
 
+    &:selected {
+        @extend %selected_item;
+    }
+
     image {
         color: #{'@placeholder_text_color'};
         transition: all duration("element") $easing;
--- a/src/widgets/_entries.scss
+++ b/src/widgets/_entries.scss
@@ -18,10 +18,6 @@ entry {
         @extend %selected_item;
     }
 
-    text selection {
-        @extend %selected_item;
-    }
-
     image {
         color: #{'@placeholder_text_color'};
         transition: all duration("element") $easing;
--- a/src/widgets/_entries.scss
+++ b/src/widgets/_entries.scss
@@ -18,6 +18,10 @@ entry {
         @extend %selected_item;
     }
 
+    text selection {
+        @extend %selected_item;
+    }
+
     image {
         color: #{'@placeholder_text_color'};
         transition: all duration("element") $easing;
--- a/src/widgets/_index.scss
+++ b/src/widgets/_index.scss
@@ -18,17 +18,23 @@ messagedialog,
 @define-color success_color @LIME_700;
 @define-color warning_color @BANANA_900;
 
+@define-color selected_bg_color #{"alpha(@accent_color_100, 0.3)"};
+@define-color selected_fg_color #{"@accent_color_900"};
+
 // Things Dazzle expects
 @define-color borders #{"" + $toplevel-border-color};
 @define-color content_view_bg #{""+ bg_color(2)};
 @define-color theme_fg_color #{"" + $fg-color};
-@define-color theme_selected_fg_color #{"" + $fg-color};
+@define-color theme_selected_bg_color #{"alpha(@accent_color_100, 0.3)"};
+@define-color theme_selected_fg_color #{"@accent_color_900"};
 @define-color wm_shadow #{rgba(black, 0.2)};
 
-*:selected,
 selection {
-    background-color: #{'alpha(@accent_color_100, 0.3)'};
-    color: #{'@accent_color_900'};
+    @extend %selected_item;
+}
+
+.gtkstyle-fallback {
+    @extend %selected_item;
 }
 
 @if $color-scheme == "dark" {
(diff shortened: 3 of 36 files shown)


danrabbit:
I’d really prefer not to use wildcard selectors if possible

lainsce:
In this case, it's needed because you wouldn't know the actual target name used by third-party apps like Firefox to display their selection, and so on (Hint: it'd bloat the files to get every single selection class used by them :grimacing:). I think this is the only wildcard we'd need to have at most. :pray:

danrabbit:
Adwaita doesn't result to using the wildcard selector, so I'm not convinced that's true

lainsce:
Adwaita instead does repetition and adds `&:selected` blocks to each widget. And there's some wacky class names there as well that I never saw, for ex. `.gtkstyle-fallback` with the `&:selected` block in it as well..

lainsce:
Oh ho, found what Adwaita really does:
```
%selected_items {
  background-color: $selected_bg_color;

  @at-root %nobg_selected_items, & {
    color: $selected_fg_color;

    @at-root %selected_items_disabled,
    &:disabled { color: mix($selected_fg_color, $selected_bg_color, 50%); }

    @at-root %selected_items_backdrop,
    &:backdrop {
      color: $backdrop_selected_fg_color;

      &:disabled { color: mix($backdrop_selected_fg_color, $selected_bg_color, 30%); }
    }
  }
}
```

They then extend this to each widget possible.

src/widgets/_entries.scss
```diff
@@ -14,6 +14,14 @@ entry {
         color: #{'@placeholder_text_color'};
     }
 
+    &:selected {
+        @extend %selected_item;
+    }
+
+    text selection {
+        @extend %selected_item;
+    }
```
danrabbit, line 23: This is not necessary since we already style `selection`

src/widgets/_index.scss
```diff
@@ -18,28 +18,36 @@ messagedialog,
...
     @define-color warning_color @BANANA_100;
 
-    @define-color selected_bg_color #{'alpha(@accent_color_700, 0.3)'};
-    @define-color selected_fg_color #{'@accent_color_100'};
-}
+    @define-color selected_bg_color #{"alpha(@accent_color_700, 0.3)"};
+    @define-color selected_fg_color #{"@accent_color_100"};
+
+    @define-color theme_selected_bg_color #{"alpha(@accent_color_700, 0.3)"};
+    @define-color theme_selected_fg_color #{"@accent_color_100"};
```
danrabbit, line 49: This is pretty redundant. We should just have these point back to `selected_bg_color` and `selected_fg_color`

src/widgets/_views.scss
```diff
@@ -29,3 +39,13 @@ scrolledwindow {
...
 }
+
+textview {
+    text selection {
+        &:focus,
+        & {
+            @extend %selected_item;
+        }
+    }
+}
```
danrabbit, line 50: Unnecessary since we already style `selection`

src/widgets/_views.scss
```diff
@@ -18,6 +18,16 @@ scrolledwindow {
...
 
+    &:selected {
+        &:focus, & {
+            @extend %selected_item;
+        }
+    }
+
+    text selection {
+        @extend %selected_item;
+    }
```
danrabbit, line 29: Unnecessary since we already style `selection`

src/_mixins.scss
```diff
@@ -192,6 +192,16 @@ $toplevel-border-color: rgba(black, 0.2);
     }
 }
 
+%selected_item {
+    @if $color-scheme == "dark" {
+        background-color: #{"@selected_bg_color"};
+        color: #{"@selected_fg_color"};
+    } @else if $color-scheme == "light" {
+        background-color: #{"@selected_bg_color"};
+        color: #{"@selected_fg_color"};
```
danrabbit, line 201: This does the exact same thing for both, so this makes no sense

lainsce:
All review points solved.

src/widgets/_index.scss
```diff
@@ -18,28 +18,35 @@ messagedialog,
...
+@define-color theme_selected_fg_color #{"@selected_fg_color"};
 @define-color wm_shadow #{rgba(black, 0.2)};
 
-@define-color selected_bg_color #{'alpha(@accent_color_100, 0.3)'};
-@define-color selected_fg_color #{'@accent_color_900'};
+selection {
+    @extend %selected_item;
+}
+
+.gtkstyle-fallback {
```
danrabbit, line 36: Can you make a comment about what this is and where it's used

src/widgets/_index.scss
```diff
@@ -18,28 +18,35 @@ messagedialog,
...
-    @define-color selected_fg_color #{'@accent_color_100'};
-}
+    @define-color selected_bg_color #{"alpha(@accent_color_700, 0.3)"};
+    @define-color selected_fg_color #{"@accent_color_100"};
 
-selection {
-    background-color: #{'@selected_bg_color'};
-    color: #{'@selected_fg_color'};
+    @define-color theme_selected_bg_color #{"@selected_bg_color"};
+    @define-color theme_selected_fg_color #{"@selected_fg_color"};
```
danrabbit, line 49: You shouldn't have to define this twice. You set it to the same thing whether it's dark or not

src/widgets/_index.scss
```diff
@@ -18,28 +18,35 @@ messagedialog,
...
 @if $color-scheme == "dark" {
     @define-color error_color @STRAWBERRY_300;
     @define-color success_color @LIME_300;
     @define-color warning_color @BANANA_100;
 
-    @define-color selected_bg_color #{'alpha(@accent_color_700, 0.3)'};
-    @define-color selected_fg_color #{'@accent_color_100'};
-}
+    @define-color selected_bg_color #{"alpha(@accent_color_700, 0.3)"};
+    @define-color selected_fg_color #{"@accent_color_100"};
```
danrabbit, line 46: Can you undo this change to quotations please

src/widgets/_views.scss
```diff
@@ -18,6 +18,12 @@ scrolledwindow {
     background-color: bg-color(1);
     color: $fg-color;
 
+    &:selected {
+        &:focus, & {
```
danrabbit, line 22: This selector doesn't really make sense to me. Can you explain what you're trying to do?

src/widgets/_entries.scss
```diff
@@ -14,6 +14,10 @@ entry {
         color: #{'@placeholder_text_color'};
     }
 
+    &:selected {
```
danrabbit, line 17: Can you give an example of when an entry would be selected?

src/widgets/_menus.scss
```diff
@@ -25,6 +25,10 @@
         background-color: #{'alpha(@fg_color, 0.15)'};
     }
 
+    &:selected {
+        @extend %selected_item;
+    }
```
danrabbit, line 30: Can you give an example of where menuitems or modelbuttons are selected?

src/widgets/_index.scss
lainsce: This is used wherever an item/widget uses GTK colors but isn't GTK proper.
Very useful when dealing with non elementary apps.

src/widgets/_index.scss
lainsce: Will fix.

src/widgets/_index.scss
lainsce: Isn't `"` more correct than `'` in Sass for this thing? The official docs say one uses `"` so I'm confused.

lainsce: https://sass-lang.com/documentation/interpolation#quoted-strings

Relevant docs

src/widgets/_views.scss
lainsce: If X is X:selected:focus or X:focus, or :selected X.

src/widgets/_entries.scss
lainsce: By default a Gtk.Entry is selected unless you remove it's focus to another widget at creation, right? This is where this comes into play.

src/widgets/_menus.scss
lainsce: Some third-party apps have this distinction for coloring. Should I remove this then?

src/widgets/_index.scss
lainsce: In another PR from this, I'll add in non-selected states to it to complete it.

src/widgets/_views.scss
lainsce: I'll rewrite this to be more concise

src/widgets/_menus.scss
lainsce: Going to remove since it's not used in elementary.

src/widgets/_entries.scss
danrabbit: You're thinking of :focus

lainsce:
Superceded by advancements in other PRs.

PR #717, opened 2020-05-26
Add basic List styles

danrabbit:
Copying the selection style here might end up being too light, but I wanted to try it

Screenshot from 2020-05-26 11 39 52@2x

* Adds `selected_bg_color` and `selected_fg_color` public variables
* Re-add the padding rule for .h4 labels
* Removes the rule for `label` because this is too specific and it breaks styles such as in the datetime indicator and calendar. We should add `color` rules to containers, but probably not labels and images themselves
* Make sure that lists have a bg-color(1), unless they're in popovers. Almost always when we use a list in a popover we don't want it to have a background. If we do, we put it inside a frame with `view` class

### DIFF
--- a/src/widgets/_lists.scss
+++ b/src/widgets/_lists.scss
@@ -7,7 +7,7 @@ list {
 
     row {
         &:selected {
-            background-color: rgba($fg-color, 0.2);
+            background-color: rgba($fg-color, 0.15);
 
             &:focus {
                 background-color: #{'@selected_bg_color'};
--- a/src/widgets/_lists.scss
+++ b/src/widgets/_lists.scss
@@ -18,5 +18,6 @@ list {
 
     .h4 {
         padding-left: rem(6px);
+        padding-right: rem(6px);
     }
 }
--- a/src/widgets/_lists.scss
+++ b/src/widgets/_lists.scss
@@ -7,7 +7,7 @@ list {
 
     row {
         &:selected {
-            background-color: rgba($fg-color, 0.15);
+            background-color: rgba($fg-color, 0.1);
 
             &:focus {
                 background-color: #{'@selected_bg_color'};
--- a/src/widgets/_lists.scss
+++ b/src/widgets/_lists.scss
@@ -1,0 +1,22 @@
+list {
+    background-color: bg-color(1);
+
+    popover & {
+        background-color: transparent;
+    }
+
+    row {
+        &:selected {
+            background-color: rgba($fg-color, 0.2);
+
+            &:focus {
+                background-color: #{'@selected_bg_color'};
+                color: #{'@selected_fg_color'};
+            }
+        }
+    }
+
+    .h4 {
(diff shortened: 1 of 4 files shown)


cassidyjames:
Yeah my initial thought is that it's too light, but I'm not sure why. Maybe because I'm so used to the saturated look for selections, and because when it's text highlighting, it makes me think of a highlighter but this doesn't.

But I'm down to give it a try for a while. :smile:

src/widgets/_lists.scss
```diff
@@ -0,0 +1,23 @@
+list {
+    background-color: bg-color(1);
+
+    popover & {
+        background-color: transparent;
+    }
+
+    row {
+        &:selected {
+            background-color: rgba($fg-color, 0.2);
```
cassidyjames, line 10: This feels quite a bit darker than the focus color. Is there a strong reason to not always use the same selected colors? It looks like Adwaita does at least.

cassidyjames: Otherwise, we could potentially try to desaturate the selected color here as well.

src/widgets/_lists.scss
```diff
@@ -0,0 +1,23 @@
...
+
+            &:focus {
+                background-color: #{'@selected_bg_color'};
+                color: #{'@selected_fg_color'};
+            }
+        }
```
cassidyjames, line 16:
```suggestion
    row {
        &:selected {
            // Effectively desaturate the background
            background-blend-mode: hue;
            background-color: #{'@selected_bg_color'};
            background-image: linear-gradient(white, white);

            &:focus {
                background-color: #{'@selected_bg_color'};
                background-image: none;
                color: #{'@selected_fg_color'};
            }
        }
```

danrabbit: This is much worse for some colors:

Screenshot from 2020-05-26 12 44 28@2x
Screenshot from 2020-05-26 12 44 38@2x
Screenshot from 2020-05-26 12 44 46@2x

src/widgets/_lists.scss
danrabbit: Yes because you need to be able to tell the difference between keyboard focus and selection. This is especially apparent in views like Files' column view

danrabbit:
@cassidyjames yeah I think it's going to end up being too light as well, but I figured it's worth trying

src/widgets/_lists.scss
```diff
@@ -0,0 +1,23 @@
+list {
+    background-color: bg-color(1);
+
+    popover & {
+        background-color: transparent;
+    }
+
+    row {
+        &:selected {
+            background-color: rgba($fg-color, 0.15);
```
cassidyjames, line 10: Okay this is a tiny change but it feels closer to the feel of the highlight colors in general. Thoughts?

```suggestion
            background-color: rgba($fg-color, 0.1);
```

PR #718, opened 2020-05-26
Toolbars: Add styles for flat actionbars

danrabbit:
Supports both `flat` and `inline-toolbar`. `Gtk.STYLE_CLASS_INLINE_TOOLBAR` is being removed from Gtk4, so we want to be able to start migrating over to `Gtk.STYLE_CLASS_FLAT`, but we probably also want to keep "legacy" support


danrabbit:
Whelp I literally already did this

PR #720, opened 2020-05-27
Typography: Bold h1 headings

cassidyjames:
Trying out something a little more modern typography-wise. I really like it, but we'll want to make sure it feels nice.

Screenshot from 2020-05-27 11 21 58
Screenshot from 2020-05-27 11 37 36Screenshot from 2020-05-27 11 21 31
Screenshot from 2020-05-27 11 21 04
Screenshot from 2020-05-27 11 20 40

### DIFF
--- a/src/_typography.scss
+++ b/src/_typography.scss
@@ -1,5 +1,6 @@
 .h1 {
     font-size: 24pt;
+    font-weight: 800;
     letter-spacing: -0.04em;
 }
 
--- a/src/_typography.scss
+++ b/src/_typography.scss
@@ -1,6 +1,6 @@
 .h1 {
     font-size: 24pt;
-    font-weight: 800;
+    font-weight: 700;
     letter-spacing: -0.04em;
 }
 


lainsce:
Nah this feels too much like Ultrabold Apple headers.

cassidyjames:
For context, we've been playing with bolder headings in different places for quite some time now, especially on the web. It's sort of part of the AppCenter/developer brand:

Screenshot from 2020-05-27 13 13 25
Screenshot from 2020-05-27 13 13 30

And we're also using much bolder headings on the blog, which has a big focus on typography:

Screenshot from 2020-05-27 13 14 30

Of course these spots are using Raleway, but we're using all-Inter in OS 6. I really like the bolder look and revisiting our typography in general. For those gut "to thicc" reactions (posted within minutes of this PR, so doubtful you've actually given it a try): give it a try for a bit. :stuck_out_tongue:  There are lighter weights we could use that are in between (and even one heavier!), but I really like Inter at the 800 weight for that bold, legible Swiss look.

Try it out for a bit.

It's also likely that we'll want to tweak other weights if we're going a little bolder in places, but I thought this was a good starting point to set the direction.

lainsce:
After a day of testing, I think 600 it better than 800. That still is bold while retaining the idea here.

cassidyjames:
600 doesn't feel right to me, but I would be okay with 700. I think I prefer 800 the most, still.

danrabbit:
Yeah 800 is too chonky imo. I'm inclined to agree that 600 is better, but I'd be willing to see if I get used to 700

4jNsY6fCVqZv:
How do the variants with 600 and 700 look in comparison?

lainsce:
After a week testing with the weight 700, I think it's better with 700.

PR #721, opened 2020-05-27
Add styles for Hdy.Avatar

danrabbit:
Screenshot from 2020-10-02 14 05 55@2x


cassidyjames (approved):
This seems to do the thing, and doesn't regress with current avatars.

PR #727, opened 2020-05-29
libHandy styles → _handy.scss

lainsce:
This PR does:

- Moves libhandy related styles to its own file
- Adds styling to `Hdy.ViewSwitcher`


lainsce:
The window style could extend from the normal one.
And libhandy is third-party so made sense to me to decouple it from the main widgets files for easy access if they externally change anything like a class, id, or something.

lainsce:
Closing this…

PR #728, opened 2020-05-29
Typography: bold dialog titles

cassidyjames:
Opening for discussion/consideration. I love this style, but also feel like the bolder titles more effectively balance out both the icon and the action buttons. It also makes it a bit easier to skim the gist of the dialog at a glance.

## 700 weight

Screenshot from 2020-06-30 15 04 12@2x

Screenshot from 2020-06-30 15 05 15@2x

Screenshot from 2020-06-30 15 07 41@2x

Screenshot from 2020-06-30 15 07 14@2x

## 800 weight

inter-800

Screenshot from 2020-05-29 12 51 27@2x

Screenshot from 2020-05-29 12 50 30@2x

Screenshot from 2020-05-29 12 54 54@2x

## master

inter-500

Screenshot from 2020-05-29 12 57 25@2x

Screenshot from 2020-05-29 12 56 43@2x

Screenshot from 2020-05-29 12 55 52@2x

### DIFF
--- a/src/_typography.scss
+++ b/src/_typography.scss
@@ -22,7 +22,7 @@ label.h4,
 }
 
 .primary {
-    font-weight: 500;
+    font-weight: 800;
     font-size: 1.22em
 }
 
--- a/src/_typography.scss
+++ b/src/_typography.scss
@@ -22,7 +22,7 @@ label.h4,
 }
 
 .primary {
-    font-weight: 800;
+    font-weight: 700;
     font-size: 1.22em
 }
 


hanaral:
It looks kinda smudgy and hard to read at anything below 1080p. I find the current master easier on the eyes too, I'd be happy to help test which thickness is easiest on the eyes if I knew the CSS node for the header text

cassidyjames:
@hanaral see the code here: it's just the `.primary` class.

lainsce:
Weight 700 feels less smudgy on LoDPI too, so I'd agree on reducing to that

cassidyjames:
Updated screenshots above to include 700-weight as well.

danrabbit:
I'm inclined to close this since the weight change to semibold here was [a suggestion from Rasmus](https://github.com/elementary/default-settings/issues/152#issuecomment-577239406)

cassidyjames:
@danrabbit [semi-bold would be `600`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#Common_weight_name_mapping), but it is still `500` unless I'm missing something.

PR #736, opened 2020-06-08
Move exported GtkCSS variables to their own file

cassidyjames:
Was from my work in https://github.com/elementary/stylesheet/compare/destructive, but pulled out. Supersedes #735 I guess. Only focuses on GtkCSS variables to keep it clear what is external vs internal API.

- Creates new `_exported.scss` file with only exported GtkCSS variables. Things that Dazzle, GTK apps, and apps using foreign drawing APIs might expect. The closest thing GTK has to a theming API.
- Renames `_mixins.scss` to `_index.css` since we're using it for variable definitions, placeholder classes, mixins, and anything else that doesn't have a better home yet.

### DIFF
--- a/src/_exported.scss
+++ b/src/_exported.scss
@@ -19,9 +19,9 @@
 @define-color placeholder_text_color #{"" + mix($fg-color, bg-color(1), $weight: 75%)};
 
 // Semantic colors
-@define-color error_color $error-color;
-@define-color success_color $success-color;
-@define-color warning_color $warning-color;
+@define-color error_color #{"" + $error-color};
+@define-color success_color #{"" + $success-color};
+@define-color warning_color #{"" + $warning-color};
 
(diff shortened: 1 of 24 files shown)


src/_exported.scss
```diff
@@ -0,0 +1,43 @@
...
+// Text and image color
+@define-color fg_color #{"" + $fg-color};
+
+// Placeholder text in entries. Doesn't respect alpha
+@define-color placeholder_text_color #{"" + mix($fg-color, bg-color(1), $weight: 75%)};
+
+// Semantic colors
+@define-color error_color $error-color;
+@define-color success_color $success-color;
+@define-color warning_color $warning-color;
```
danrabbit, line 24:
```suggestion
@define-color error_color #{"" + $error-colo};
@define-color success_color #{"" + $success-color};
@define-color warning_color #{"" + $warning-color};
```

PR #742, opened 2020-06-12
Initial focus styles

danrabbit:
very github-like focus style. Trying to find something that works for many different types of widgets

Something to play with is outline style and where it lands on the z-axis


cassidyjames:
I do like this in general! I think the highlight "glow" could be a little softer on the edges, but otherwise it's a great start.

PR #744, opened 2020-06-12
Style emoji picker

danrabbit:


src/widgets/_emojipicker.scss
```diff
@@ -0,0 +1,58 @@
...
+        font-weight: 700;
+        //Intentionally off-by-one
+        margin-left: rem(7px);
+        margin-right: rem(7px);
+        opacity: 0.8;
+    }
+
+    .emoji {
+        min-height: 48px;
+        min-width: 48px;
```
cassidyjames, line 18: This seems pretty spaced out, reducing the amount of emoji that can be seen without scrolling. What about 36px here?

```suggestion
        min-height: 36px;
        min-width: 36px;
```

src/widgets/_emojipicker.scss
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    // Emoji are labels too, so we have to be specific
+    box.vertical > label {
+        font-weight: 700;
+        //Intentionally off-by-one
+        margin-left: rem(7px);
+        margin-right: rem(7px);
+        opacity: 0.8;
+    }
```
cassidyjames, line 14: If you're trying to match `.h4` styling, just extend it to ensure it stays in sync.

```suggestion
    box.vertical > label {
        @extend .h4;

        // Intentionally off-by-one
        margin-left: rem(7px);
        margin-right: rem(7px);
    }
```

danrabbit: woah didn't know you could do that. That is extremely useful

PR #745, opened 2020-06-12
Button: Add spacing between images and labels

danrabbit:
When a button contains both an image and a label, make sure we add column spacing

### DIFF
--- a/src/widgets/_buttons.scss
+++ b/src/widgets/_buttons.scss
@@ -100,4 +100,9 @@ button {
     label {
         font-weight: 500;
     }
+
+    &:dir(ltr) image + label,
+    &:dir(rtl) label + image {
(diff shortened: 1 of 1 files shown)


cassidyjames:
In Widget Factory, this looks like too much spacing to me. Is the styling there just weird, or do you have another example of where this can be seen?

## master

screenshot-0

## this branch

screenshot-1

danrabbit:
This happens in the old stylesheet too, it looks like in the widget factory they're manually packing their own box with spacing:

Screenshot from 2020-06-16 10 46 16@2x

Here's a button that just uses the `always-show-image` property:

before:
Screenshot from 2020-06-16 10 47 14@2x

After:
Screenshot from 2020-06-16 10 47 53@2x

PR #747, opened 2020-06-15
Typography: Color links with accent color

cassidyjames:
Also adds underline on hover. See Onboarding or the About plug for some examples.


cassidyjames:
@danrabbit do you think we should just always underline them? With `.accent` support, I'm concerned we're defaulting to it not being clear what is a link and what isn't without hovering over them. It's a little less pretty to always underline, but we also don't use link buttons a _ton_ so it might be fine.

PR #752, opened 2020-06-19
Window: support rounded class

danrabbit:


src/widgets/_windows.scss
```diff
@@ -69,6 +69,14 @@ window:not(.popup):not(.menu) {
         }
     }
 
+    &.rounded {
+        border-radius: 0 0 rem(6px) rem(6px);
```
Exalm, line 73: Interesting, I'd expect rounded to be rounded on all sides.

danrabbit: Yeah you’d think that but what you end up with is rounded corners beneath your titlebar :p

Exalm: Ah right. 🤦️

PR #754, opened 2020-06-19
Typography: Add keycap style

danrabbit:
as in the shortcut overlay

Ideally, shouldn't change menus

### DIFF
--- a/src/_typography.scss
+++ b/src/_typography.scss
@@ -44,7 +44,7 @@ label.h4,
         linear-gradient(
             to bottom,
             rgba(white, 0),
-            #{'alpha(@highlight_color, 0.2)'},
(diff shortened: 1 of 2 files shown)


src/_typography.scss
```diff
@@ -38,6 +38,35 @@ label.h4,
     opacity: 0.75;
 }
 
+.keycap {
+    background-color: bg-color(2);
+    background-image:
+        linear-gradient(
+            to bottom,
+            rgba(white, 0),
+            #{'alpha(@highlight_color, 0.2)'},
```
cassidyjames, line 47:
```suggestion
            #{'alpha(@highlight_color, 0.2)'}
```

Issue #573, opened 2020-01-03
The line between the source list and folder view sometimes disappears


ryonakano:
![peek 2018-09-14 14-49](https://user-images.githubusercontent.com/26003928/45532081-a0771d00-b82d-11e8-8d3d-f7e4d5c255ac.gif)

It may be due to my system environment, but the line between the source list and folder view sometimes disappears.
I'm on an Core i7-2630QM with NVIDIA GeForce GT 540M. I installed NVIDIA binary driver version 340.106 for GPU from AppCenter.

msmaldi:
I have same Problem, I use Nvidia, new themes have bug too.
Change rendering mode to image resolve, is possible change for all applications permant?

PR #756, opened 2020-06-23
Panes: fix extra margin

danrabbit:
Fixes the weird extra gap after panes. Looks like maybe a rounding error

Also may address #573 maybe?

### DIFF
--- a/src/widgets/_panes.scss
+++ b/src/widgets/_panes.scss
@@ -12,13 +12,11 @@ paned.horizontal > separator {
         background-image:
             linear-gradient(
                 to right,
-                #{'@menu_separator'} 1px,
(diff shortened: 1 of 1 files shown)


src/widgets/_panes.scss
```diff
@@ -12,13 +12,11 @@ paned.horizontal > separator {
         background-image:
             linear-gradient(
                 to right,
-                #{'@menu_separator'} 1px,
-                rgba(white, 0.01) 1px,
-                rgba(white, 0.01) 2px,
```
danrabbit, line 17: This is never shown because of the way z-index works in gtk. The pane is under the view next to it

PR #758, opened 2020-06-30
Menus: reset labels and images

danrabbit:
Fixes issues with white text and images in indicators as well as bold menu items from menu buttons


hanaral:
Maybe a regression? I still get this every so often:
Screenshot from 2021-01-08 11-40-54

PR #760, opened 2020-07-06
Scales: add temp and warmth scales

### LINKED ISSUE #792: Missing warmth and temperature styles
Used in Photos and I think Night Light? See https://github.com/elementary/granite/issues/306.

danrabbit:
Fixes #792 
As used in Nightlight


danrabbit:
@cassidyjames It looks like we took the white out of the middle in Photos and are actually using Yellow there instead of Orange. So I've updated this to match the style we're shipping there

PR #761, opened 2020-07-06
Style expanders

danrabbit:


src/widgets/_expanders.scss
```diff
@@ -0,0 +1,30 @@
+expander arrow,
+.expander,
+.category-expander {
+    min-height: 16px;
+    min-width: 16px;
+    transition: all duration("element") $easing;
```
cassidyjames, line 6: It's possible because I'm on the Pinebook Pro, but I'm not seeing an animation here at all. I'll check on more powerful hardware, though.

cassidyjames: @danrabbit so even on substantially more powerful hardware, I'm only just _sometimes_ seeing this animation. I wonder if GTK doesn't like the easing we're using and so is deciding to skip it in favor of performance? But I suppose that can be addressed in a branch that affects the easing itself.

PR #762, opened 2020-07-09
Dim images and labels when insensitive

### LINKED ISSUE #772: Insensitive image-button in GtkToolBar doesn't feel insensitive
Capture d'écran du 2020-07-21 17 38 47@2x
In this screenshot, all the elements of the toolbar from the reply button to the end are insensitive. The lack of the usual image transparency makes it hard to notice it.

danrabbit:
Fixes #772


src/_typography.scss
```diff
@@ -38,6 +38,10 @@ label.h4,
     opacity: 0.75;
 }
 
+label:disabled {
+    opacity: 0.5;
```
lainsce, line 42: I feel this gives less contrast for an insensitive label under Greenfield... maybe 0.66 at least? :)

PR #763, opened 2020-07-12
Notebook Tabs: move padding to grid

### LINKED ISSUE #523: DynamicNotebook: Padding and border around Gtk.EventBox
Conversation from https://github.com/elementary/granite/pull/295:

---

@andreasfelix Definitely better than before! It still happens when I double-wheelclick on the border between two tabs, though. Do you think this is can/should be fixed within the same PR?

_Originally posted by @jlnr in https://github.com/elementary/granite/pull/295#issuecomment-508928380_

---

Unfortunately I think this is a different Issue. I have highlighted the size of the `GranitsWidgetsTab` and logged the enter and leave events:

Border_event

As one can see, the size of the `GranitsWidgetsTab` is slightly smaller than its visual size. This can also be seen at the event log: The signal get emitted when entering/leaving the blue highlighted area. When clicking on a Tab outside of this are, the signal gets directly passed to `DynamicNotebook` which leads to the restoring of the last closed tab.

I think this is related to the `padding` and `border` in the style sheet:

https://github.com/elementary/stylesheet/blob/ec23fec4a4bcd6577d2d2039eb24ae9bc9364577/elementary/gtk-3.0/gtk-widgets.css#L1670-L1688
Maybe it could be solved by changing `notebook tab` to `notebook tab widgets grid`, so the padding is within the `Gtk.EventBox`.

_Originally posted by @andreasfelix in https://github.com/elementary/granite/pull/295#issuecomment-508934896_

---

andreasfelix:
Fixes #523

Currently there is a padding around the `Granite.Widgets.Tab`. So clicks within the 3px edge of the tab are not registered and instead ripple through to the `Granite.DynamicNotebook`. This can lead to the problem, that when trying to close a tab with <kbd>Middle</kbd>  or <kbd>Three-Finger</kbd><kbd>Click</kbd> an old tab is restored instead.

**Before** (EventArea of `GraniteWidgetsTab` )
tab-before

**After** (EventArea of `GraniteWidgetsTab` )
tab-after

### DIFF
--- a/src/widgets/_notebooks.scss
+++ b/src/widgets/_notebooks.scss
@@ -27,7 +27,6 @@ tab {
     margin: rem(3px) 0;
     // Not scaled because it's based on icon size
     min-width: 24px;
-    padding: rem(3px);
 
     &:hover:not(:checked) {
         background-color: bg-color(4);
@@ -43,5 +42,9 @@ tab {
         &:backdrop {
             background-color: bg-color(2);
         }
+    }
+
+    > widget > grid {
+        padding: rem(3px);
     }
 }


andreasfelix:
This also effects the context menu of the tabs. When right-clicking the within the 3px padding of a tab, instead of the context menu of the the tab the context menu of the tab-bar is shown.

PR #764, opened 2020-07-13
Titles & Subtitles

lainsce:
Porting over missing title and subtitle styling, while also updating it for the new stylesheet.

Looks like this:
Screenshot from 2020-07-12 23 05 44

### DIFF
--- a/src/widgets/_toolbars.scss
+++ b/src/widgets/_toolbars.scss
@@ -15,6 +15,11 @@ actionbar {
     &.flat{
         @extend .inline-toolbar;
     }
+    &.title {
+        font-size: 11pt;
+        font-weight: bold;
+        padding: 0 12px;
+    }
 }
 
 .search-bar {
--- a/src/widgets/_toolbars.scss
+++ b/src/widgets/_toolbars.scss
@@ -15,9 +15,15 @@ actionbar {
     &.flat{
         @extend .inline-toolbar;
     }
+
     &.title {
         font-size: 11pt;
         font-weight: bold;
+        padding: 0 12px;
+    }
+
+    &.subtitle {
+        font-size: 9pt;
         padding: 0 12px;
     }
 }
--- a/src/widgets/_toolbars.scss
+++ b/src/widgets/_toolbars.scss
@@ -19,7 +19,7 @@ actionbar {
     &.title {
         font-size: 11pt;
         font-weight: 600;
-        padding: 0 12px;
+        padding: 0 rem(12px);
     }
 
     &.subtitle {
--- a/src/widgets/_toolbars.scss
+++ b/src/widgets/_toolbars.scss
@@ -15,17 +15,6 @@ actionbar {
     &.flat{
         @extend .inline-toolbar;
     }
-
-    &.title {
-        font-size: 11pt;
-        font-weight: 600;
-        padding: 0 rem(12px);
-    }
-
-    &.subtitle {
-        font-size: 9pt;
-        padding: 0 rem(12px);
-    }
 }
 
 .search-bar {
--- a/src/widgets/_toolbars.scss
+++ b/src/widgets/_toolbars.scss
@@ -24,7 +24,7 @@ actionbar {
 
     &.subtitle {
         font-size: 9pt;
-        padding: 0 12px;
+        padding: 0 rem(12px);
     }
 }
 
--- a/src/widgets/_toolbars.scss
+++ b/src/widgets/_toolbars.scss
@@ -18,7 +18,7 @@ actionbar {
 
     &.title {
         font-size: 11pt;
-        font-weight: bold;
+        font-weight: 600;
         padding: 0 12px;
     }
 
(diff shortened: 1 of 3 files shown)


src/widgets/_windows.scss
```diff
@@ -170,10 +170,16 @@ window:not(.popup):not(.menu) {
...
 
         .title {
+            font-size: 9pt;
             font-weight: 600;
             text-shadow: $shadow
         }
 
+        .subtitle {
+            opacity: 0.75;
+            font-size: 85%;
```
danrabbit, line 180: This is larger than what we had before
```suggestion
            font-size: 80%;
```

src/widgets/_windows.scss
```diff
@@ -170,10 +170,16 @@ window:not(.popup):not(.menu) {
         background-color: #{"@color_primary"};
 
         .title {
+            font-size: 9pt;
```
danrabbit, line 173: Is this necessary?

src/widgets/_toolbars.scss
```diff
@@ -15,6 +15,17 @@ actionbar {
     &.flat{
         @extend .inline-toolbar;
     }
+
+    &.title {
+        font-size: 11pt;
+        font-weight: bold;
```
danrabbit, line 21: Let's be more specific here to match what we've done in window titles
```suggestion
        font-weight: 600;
```

src/widgets/_toolbars.scss
```diff
@@ -15,6 +15,17 @@ actionbar {
...
+        padding: 0 12px;
```
danrabbit, line 22: Don't forget rems
```suggestion
        padding: 0 rem(12px);
```

src/widgets/_toolbars.scss
```diff
@@ -15,6 +15,17 @@ actionbar {
...
+    }
+
+    &.subtitle {
+        font-size: 9pt;
+        padding: 0 12px;
```
danrabbit, line 27:
```suggestion
        padding: 0 rem(12px);
```

src/widgets/_toolbars.scss
```diff
@@ -15,6 +15,17 @@ actionbar {
...
     }
+
+    &.title {
+        font-size: 11pt;
+        font-weight: bold;
+        padding: 0 12px;
+    }
+
+    &.subtitle {
+        font-size: 9pt;
```
danrabbit, line 26: is this necessary since that's the default size?

src/widgets/_toolbars.scss
```diff
@@ -15,6 +15,17 @@ actionbar {
     &.flat{
         @extend .inline-toolbar;
     }
+
+    &.title {
+        font-size: 11pt;
```
danrabbit, line 20: does using `pt` units scale when the user changes their font size?

Also do you know where this is used? Is there a reason these are large than the window title? That seems a little odd

lainsce: `pt` is meant to be scalable in the context of fonts only.

Also, some apps add labels with a title/subtitle styling to actionbars, and these are larger to accomodate the styling seen in their screenshots.

src/widgets/_windows.scss
lainsce: It's necessary since label size might change with the user prefence, and enforcing a size using `pt` ensures it scales properly.

src/widgets/_toolbars.scss
lainsce: As said in the title comment, it's necessary since label size might change with the user prefence, and enforcing a size using `pt` ensures it scales properly.

src/widgets/_windows.scss
danrabbit: Can you elaborate on what you mean by this? As far as I can tell, this seems to have no visible effect

src/widgets/_toolbars.scss
danrabbit: Can you elaborate on what you mean by "some apps"? A specific example would help here.

I'm not really worried about breaking screenshots since this is an entirely new stylesheet and it will break screenshots regardless

lainsce: Yeah, I think that it's maybe best to not write for edge cases then. An example was Synaptic; but since it's meant to be a cleaner stylesheet, I'll remove this.

src/widgets/_windows.scss
lainsce: It's basically "more correct" measurement unit wise. Points are scaled automatically to a whole number of physical pixels in order to accommodate for screen size, pixel density and typical viewing distance, so it means it's the better measurement unit to use.

lainsce: And third party apps might want to change their titles & subtitles sizes, so this ensures the starting point has a default.

danrabbit: Can you post a screenshot showing the difference between including this line versus not including it?

lainsce: The changes most visible is with the red highlights in both here:

`no-line`
no-line

`with-line`
with-line

Looking at it, it's visible the font also doesn't have wild kerning anymore (check it out with glimpse, make `with-line` semi-opaque above `no-line` and see that both don't match perfectly. The `with-line` looks better then.)

PR #765, opened 2020-07-13
Windows: Fix radius and background for tiled and maximized windows

danrabbit:


cassidyjames:
Eh but I guess this is still better than in `master` since it fixes the inside corners. I'll open a new issue and we can address it later.

PR #769, opened 2020-07-14
Windows: Set radius of tiled decoration overlay

danrabbit:
Doesn't fix the radius of the actual window, just the highlight overlay


cassidyjames:
@danrabbit what am I looking at here? i.e. what apps in what contexts?

danrabbit:
@cassidyjames this is for tiled apps using Hdy.Window. It fixes the radius of the little highlight. Much easier to see in dark style

cassidyjames:
So from what I can see, this makes the window corners non-rounded when maximized. It doesn't seem to affect when they're half-tiled. I'm fine merging if that's what it's expected to do @danrabbit

danrabbit:
I don't remember what this is supposed to do honestly lol

cassidyjames:
@danrabbit in that case, I'm gonna close this until it gets revisited. I think the stylesheet and Handy have both changed since this was originally written, and there's no use leaving it open if neither of us know what it's trying to do.

PR #774, opened 2020-07-22
Ship indicator-less themes

lainsce:
This will be necessary to make the option of hiding indicators via the Appearance plug


cassidyjames:
This won't work with the dark style switching, because Plank just looks for a `dock-dark.theme` file in response to the user preference.

lainsce:
We just need to make it be smarter then, looking for the specific variant of Plank theming we want. It's easier than changing Plank code.

cassidyjames:
Ah you're right. https://github.com/elementary/dock/pull/45

lainsce:
Closing in favor of new logic on https://github.com/elementary/dock/pull/59

PR #780, opened 2020-08-08
General: Increase contrast overall in Dark Styling

lainsce:
Makes buttons, switches, and some background contexts have increased contrast in the Dark Styling.

Exs.
a
a (1)
a (2)

Edit:

Added a new SCSS thing for inset-backgrounds, using it on pressed states:
a


cassidyjames:
When we've talked about this before, we mentioned it might make more sense to alter the dark/BLACK palette itself instead of special casing the dark style everywhere.

src/_palette.scss
```diff
@@ -53,8 +53,8 @@ $COCOA_500: #715344;
 $COCOA_700: #57392d;
 $COCOA_900: #3d211b;
 
-$SILVER_100: #fafafa;
-$SILVER_300: #d4d4d4;
+$SILVER_100: #fcfcfc;
+$SILVER_300: #e3e3e3;
```
danrabbit, line 57: If you want to propose changes to the light style, please do that in a separate branch

src/widgets/_buttons.scss
```diff
@@ -21,7 +21,11 @@ button {
     &.raised.image-button,
     & {
         @extend %outset-background;
-        background-color: bg-color(2);
+        @if $color-scheme == "light" {
+          background-color: bg-color(2);
+        } @else if $color-scheme == "dark" {
+          background-color: mix($BLACK_300, $BLACK_500, $weight: 70%);
+        }
```
danrabbit, line 28: This is going to make it more difficult to keep things consistent. Instead of doing this, you should change `bg-color(2)` to return this value for the dark style

lainsce: If I do that, then everything at level 2 of bg-color will be as raised as the buttons, defeating some of the purpose of the PR. What should I do then?

src/_palette.scss
lainsce: Will revert this.

lainsce:
> When we've talked about this before, we mentioned it might make more sense to alter the dark/BLACK palette itself instead of special casing the dark style everywhere.

I thought it made more sense to style things separately so as to refine each widget to its intended depth/contrast, instead of a catch-all which would only serve to flatten everything; e.g. a button shouldn't have that same depth as the view it's in, so it rises up.

cassidyjames:
Hm, so the actual underlying issue is that buttons have the same level as backgrounds. Which seems to work alright in the light style, because the shadows provide the apparent depth. But I agree it doesn't work as well in the dark style. @danrabbit would it be crazy to introduce another depth/color here in between backgrounds and inputs that allowed us to increase contrast of buttons compared to backgrounds?

cassidyjames:
This is superseded by adding a new input level, as in #802.

PR #800, opened 2020-09-01
Toolbars: Style flat actionbars and their buttons

danrabbit:
Screenshot from 2020-09-01 11 42 19@2x

### DIFF
--- a/src/widgets/_toolbars.scss
+++ b/src/widgets/_toolbars.scss
@@ -7,6 +7,22 @@
             rgba(black, 0.07) 1px,
             transparent rem(3px)
         );
+
+    .sidebar & {
+        background-color: mix(bg-color(3), bg-color(4), 50%);
+        background-image:
+            linear-gradient(
+                to bottom,
+                rgba(black, 0.3),
+                rgba(black, 0.04) 1px,
+                transparent rem(3px)
+            );
+        padding: rem(6px);
+
+        button {
+            @extend .flat;
+        }
+    }
 }
 
 actionbar {
--- a/src/widgets/_toolbars.scss
+++ b/src/widgets/_toolbars.scss
@@ -17,7 +17,7 @@
                 rgba(black, 0.04) 1px,
                 transparent rem(3px)
             );
-        padding: rem(6px);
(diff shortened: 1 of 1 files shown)


cassidyjames:
@danrabbit this looks really nice here in Files, but should it look this way everywhere? Right now we're pretty inconsistent between inline toolbars in sidebars and in other contexts. Compare:

### Code

Screenshot from 2020-09-01 15-31-23@2x

### Applications Plug

Screenshot from 2020-09-01 15-31-51@2x


### Wallet Plug

Screenshot from 2020-09-01 15-32-40@2x

danrabbit:
I think what we had discussed before was adding labels to these buttons in inline toolbars since we had consistent user feedback that it was difficult to understand how to add items to these lists

src/widgets/_toolbars.scss
```diff
@@ -7,6 +7,22 @@
...
+    .sidebar & {
+        background-color: mix(bg-color(3), bg-color(4), 50%);
+        background-image:
+            linear-gradient(
+                to bottom,
+                rgba(black, 0.3),
+                rgba(black, 0.04) 1px,
+                transparent rem(3px)
+            );
+        padding: rem(6px);
```
cassidyjames, line 20: I think this is what's really chonking it up for me. How about:

```suggestion
        padding: rem(3px);
```

cassidyjames:
@danrabbit sorry—no that I know, I just meant the styling here in the stylesheet. The sidebar ones seem waaay chonkier. I think scaling the padding back a little bit helps, but we should see if we want  to style the other inline toolbars with similar padding, and if so, if we should do that in the same place.

src/widgets/_toolbars.scss
danrabbit: Can we keep the horizontal padding at 6px so that the icons line up with the ones in the sidebar items?

cassidyjames: @danrabbit that works in Files but it doesn't line up with anything in Code. Not sure how much that matters. But yeah `3px 6px` is still better than 6px all around.

PR #802, opened 2020-09-03
Add a bg level for inputs

### LINKED ISSUE #671: Dark palette is too low-contrast
We discussed this on Slack but I figured I'd report here so we don't lose track of it. It seems that the colors used in the dark palette, at least in `bg-color()`, are too low contrast. The end result is that all dark style widgets don't contrast from one another as much as the light style ones, at least perceptually. We could fix this by tweaking the values used by `bg-color`, or by tweaking the `BLACK` palette itself.

danrabbit:
Supersedes #780 
Fixes #671 

Introduces a new bg level 0 for inputs. This is used for buttons, switch handles, sliders, checkboxes, etc

## BEFORE
Screenshot from 2020-09-03 13 24 14@2x

## AFTER
Screenshot from 2020-09-03 13 21 16@2x

This also darkens the view color which makes apps like Files not so ugly:

## BEFORE
Screenshot from 2020-09-03 13 28 23@2x

## AFTER
Screenshot from 2020-09-03 13 27 51@2x

You'll also notice no significant impact on the light style:

Screenshot from 2020-09-03 13 26 48@2x
Screenshot from 2020-09-03 13 26 22@2x

### DIFF
--- a/src/_index.scss
+++ b/src/_index.scss
@@ -24,8 +24,11 @@ $warning_color: $BANANA_900;
 
 @function bg-color($level) {
     @if $color-scheme == "light" {
-        // Views and inputs
-        @if $level == 1 {
+        // Inputs
+        @if $level == 0 {
+            @return $SILVER_100;
+        // Views
+        } @else if $level == 1 {
             @return white;
         // Background
         } @else if $level == 2 {
@@ -38,8 +41,10 @@ $warning_color: $BANANA_900;
             @return $titlebar-color;
         }
     } @else if $color-scheme == "dark" {
-        @if $level == 1 {
-            @return $BLACK_300;
+        @if $level == 0 {
+            @return mix($BLACK_300, $BLACK_500, $weight: 50%);
+        } @else if $level == 1 {
+            @return mix($BLACK_300, $BLACK_500, $weight: 25%);
         } @else if $level == 2 {
             @return $BLACK_500;
         } @else if $level == 3 {
--- a/src/widgets/_buttons.scss
+++ b/src/widgets/_buttons.scss
@@ -21,8 +21,14 @@ button {
     &.raised.image-button,
     & {
         @extend %outset-background;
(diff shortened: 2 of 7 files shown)


4jNsY6fCVqZv:
@danrabbit When I look at the two Files screenshots: The mediakeys.svg thumbnail is not very well recognizable for me. Somehow it reminds me of the strong white in the light theme, where I had a similar perception. Have you already thought about it?

danrabbit:
@4jNsY6fCVqZv please open an issue against Files. That doesn't have anything to do with this pull request

cassidyjames:
@danrabbit so I know it'll require a bit more find/replacing, but I do think it would be better to keep the "level" more semantically-numbered; it feels weird to me that buttons and things that extrude from the background are a "lower" level.

Another option would be to drop the numbering, and instead rely on strings for semantics. That's _probably_ more clear anyway as you don't have to refer back to the `bg-level` function to remember which level maps to which thing. If this is the way you'd want to go, merging as-is would be fine, and then I can submit a follow-up to switch it to strings.

cassidyjames:
Something else I'm noticing is that it's very difficult now to distinguish between a sensitive and insensitive input in the dark style… buttons look much higher contrast, but I think we'll need to address this, too.

danrabbit:
@cassidyjames yeah changing to strings would probably be better. Like we discussed before it's not exactly a scale of height so much as it is a scale of attention or focus I guess. So 0 makes sense here as being closest, then 1 for views, 2 for backgrounds, 3 for sidebars, 4 for toolbars. But yeah if we could come up with names that would probably be clearest

There's a lot of things that can be done with insensitive states etc, but I think that's outside the scope of this PR

cassidyjames (approved):
@danrabbit in that case, I will merge as-is and we can iterate. 😎🎉

danrabbit:
Yay! Grassy ass

PR #820, opened 2020-09-11
Move border color to index

danrabbit:
More consistent border color, especially in dark theme:

## BEFORE
Screenshot from 2020-09-11 11 33 51@2x
Screenshot from 2020-09-11 11 34 10@2x

## AFTER
Screenshot from 2020-09-11 11 33 25@2x
Screenshot from 2020-09-11 11 32 26@2x


cassidyjames (approved):
I have some concerns about the reduction of contrast, but I like the consistency, and we can see how this feels and if we want to tweak later.

PR #824, opened 2020-09-12
Dialog gets resized on focus/unfocus

### LINKED ISSUE #804: Dialog gets resized on focus/unfocus
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Describe the bug
Gtk.Dialogs are resized when they get or lose focus. I assume this is caused by the following code:
https://github.com/elementary/stylesheet/blob/be9ecefc174e9286ffeeab4e1b850cc7e758a1e7/src/widgets/_windows.scss#L235-L253

When overriding the styles with `dialog decoration { box-shadow:none; }` the issue goes away (together with any shadows the dialog had, obviously). Also for normal windows the size of the box shadow is hardcoded for probably the same reason.

## Expected behavior
The dialog should not change in size when gaining / losing focus.

## Platform Information
 - OS: elementaryOS 6

sebastianlay:
Fixes #804 and improves the readme a little bit.

And for your viewing pleasure:

comparison

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,11 @@ Run `meson` to configure the build environment. To install, use `ninja install`.
     cd build
     ninja install
 
+Run `ninja clean` before building the stylesheets a second time. Otherwise the already built stylesheets will not be overriden.
+
+    ninja clean
+    ninja install
+
 ## Hacking
 
 Apps may need to be restarted or the system stylesheet will need to be changed for your changes to take effect.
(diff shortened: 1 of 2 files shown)


danrabbit (dismissed):
Can you add a comment about the reason for the transparent box shadow? I know why you need it, but 5 years from now we might forget :)

sebastianlay:
Sure, I copied the comment from above.

cassidyjames (approved):
Thanks for adding the comment! I can confirm this fixes the dialog resizing issue for me.

PR #833, opened 2020-09-15
Badges

danrabbit:
Add a badge style that will work with applications menu and AppCenter

Also support sidebar badges as in Mail


cassidyjames (approved):
:shipit:

PR #839, opened 2020-09-28
Windows: Handle dialog action area margin

### LINKED ISSUE #281: Set dialog action area margin
`get_action_area` is deprecated in Gtk.Dialog and what we're currently using that for is to set margins. Since we can do that in CSS now, we probably should anyways and then we can fix this deprecation warning in all of our dialogs.

The painful thing is that this affects literally every dialog. Luckily we've ported a lot of dialogs to Granite.MessageDialog which should save some pain, but there are still many many other dialogs :(

There are a couple of style classes here: `dialog-action-box` and `dialog-action-area`

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/55721238-set-dialog-action-area-margin?utm_campaign=plugin&utm_content=tracker%2F45189256&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F45189256&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #281 

This is gonna break a bunch of stuff, but might as well get it over with.

Goes with https://github.com/elementary/granite/pull/374

### DIFF
--- a/src/widgets/_windows.scss
+++ b/src/widgets/_windows.scss
@@ -272,4 +272,18 @@ messagedialog {
             color: transparent;
         }
     }
+
+    .dialog-vbox {
+        // Remove the 2px border_width from Gtk.Dialog. Intentionally not in ems
+        margin: -2px;
+    }
+
+    // These should should in 12px borders around buttons
+    .dialog-action-area {
+        // Compensate for off-by-one
+        margin: rem(7px);
+        margin-bottom: rem(8px);
+        // Double space between content and actions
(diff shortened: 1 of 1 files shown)


cassidyjames (approved):
Can confirm this breaks dialog spacing in a bunch of places. :stuck_out_tongue: But it looks good with the linked Granite branch.

PR #840, opened 2020-09-28
Remove separators from StackSidebar

danrabbit:
As seen in GraniteDemo. I don't want seps in my stacksidebar


cassidyjames:
This doesn't seem to get rid of them for me…

cassidyjames (changes requested):
It looks like `separator.horizontal` from `_separators.scss` is higher specificity than `stacksidebar separator` here. So we probably need to add `.horizontal` here, and then I wonder if this should live in `_separators.scss` alongside the other separator styles, since it's overriding them?

cassidyjames (approved):

PR #841, opened 2020-09-28
Scales: Fix size and contrast in dark style

### LINKED ISSUE #656: Scales are too low-contrast in the dark style
In the old stylesheet they were much more contrasty and it was nice. It looks like for fill color we want to use some variant of fg-color instead of shading:

Screenshot from 2020-05-07 16 44 55@2x

danrabbit:
Fixes #656 

## BEFORE
Screenshot from 2020-09-28 13 09 50@2x
Screenshot from 2020-09-28 13 10 08@2x

## AFTER
Screenshot from 2020-09-28 13 08 40@2x
Screenshot from 2020-09-28 13 08 29@2x

### DIFF
--- a/src/_index.scss
+++ b/src/_index.scss
@@ -159,8 +159,13 @@ $warning_color: $BANANA_900;
         $text-size: $text-size * 1px;
     }
 
-    // Workaround GTK clamping instead of rounding up
-    @return ($pixels / $text-size * 1rem) + 0.000000001rem;
+    @if ($pixels > 1) {
+        // Workaround GTK clamping instead of rounding up
+        @return ($pixels / $text-size * 1rem) + 0.000000001rem;
+    } @else {
+        // Workaround GTK clamping instead of rounding up
+        @return ($pixels / $text-size * 1rem) - 0.000000001rem;
+    }
 }
 
 // Background for elements with outset style like headerbars, buttons, checkboxes, etc
--- a/src/_index.scss
+++ b/src/_index.scss
@@ -159,7 +159,7 @@ $warning_color: $BANANA_900;
         $text-size: $text-size * 1px;
     }
 
-    @if ($pixels > 1) {
+    @if ($pixels > 0) {
         // Workaround GTK clamping instead of rounding up
         @return ($pixels / $text-size * 1rem) + 0.000000001rem;
     } @else {
--- a/src/_index.scss
+++ b/src/_index.scss
@@ -7,6 +7,9 @@ $fg-color: $BLACK_500;
 // Outset box shadow or border color on toplevel elements like windows, menus, popovers
 $toplevel-border-color: rgba(black, 0.2);
 
+// Outset box shadow or border color on inputs like buttons, entries, checkboxes
+$border-color: rgba(black, 0.2);
+
 // Semantic colors
 $error_color: $STRAWBERRY_500;
 $success_color: $LIME_700;
@@ -16,6 +19,7 @@ $warning_color: $BANANA_900;
     $titlebar-color: mix($BLACK_500, $BLACK_700, $weight: 70%);
     $fg-color: white;
     $toplevel-border-color: rgba(black, 0.75);
+    $border-color: rgba(black, 0.3);
 
     $error-color: $STRAWBERRY_300;
     $success-color: $LIME_300;
--- a/src/_index.scss
+++ b/src/_index.scss
@@ -59,6 +59,9 @@ $warning_color: $BANANA_900;
     }
 }
 
+// Disabled text, images, other foreground elements
+$insensitive-fg-color: mix($fg-color, bg_color(2), $weight: 50%);
+
 @function shadow($level) {
     @if $color-scheme == "light" {
         @if $level == 1 {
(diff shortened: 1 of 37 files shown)


src/widgets/_scales.scss
```diff
@@ -4,10 +4,21 @@ scale {
...
-        background-color: rgba(black, 0.45);
+        background-color: rgba($fg-color, 0.7);
+
+        box-shadow:
+            inset 0 0 0 1px rgba(black, 0.3),
+            inset 0 0 0 2px rgba(black, 0.03);
+
+        @if $color-scheme == "dark" {
+            background-clip: padding-box;
+            border: 1px solid transparent;
```
cassidyjames, line 15: By adding a border here—even transparent—it means widgets shift around when shifting between the light and dark style; see the Style Classes/Constants page in Granite Demo. We should probably always include the border to prevent affecting layout between the two styles.

cassidyjames (changes requested):

src/_index.scss
```diff
@@ -159,8 +159,13 @@ $warning_color: $BANANA_900;
...
 
-    // Workaround GTK clamping instead of rounding up
-    @return ($pixels / $text-size * 1rem) + 0.000000001rem;
+    @if ($pixels > 1) {
+        // Workaround GTK clamping instead of rounding up
+        @return ($pixels / $text-size * 1rem) + 0.000000001rem;
+    } @else {
+        // Workaround GTK clamping instead of rounding up
+        @return ($pixels / $text-size * 1rem) - 0.000000001rem;
+    }
```
cassidyjames, line 168: Am I reading this right? It looks like you're returning the same thing, regardless?

danrabbit: No one is `-` and one is `+`. We were always doing `+` before which if you pass in -6 and then +0.00001 that's not the same as +6 and then +0.000001

hanaral:
It looks kinda borked at the edges for me:
image
and
image

src/_index.scss
cassidyjames: Ah, make sense. Shouldn't this be based around `0`, not `1`, then?

src/_index.scss
```diff
@@ -159,8 +159,13 @@ $warning_color: $BANANA_900;
         $text-size: $text-size * 1px;
     }
 
-    // Workaround GTK clamping instead of rounding up
-    @return ($pixels / $text-size * 1rem) + 0.000000001rem;
+    @if ($pixels > 1) {
```
cassidyjames, line 162: If this is supposed to be a switch for negative numbers, then this makes more sense… right?

```suggestion
    @if ($pixels > 0) {
```

src/widgets/_scales.scss
```diff
@@ -131,13 +142,15 @@ scale {
         highlight.top,
         fill.top {
             border-radius: rem(12px) 0 0 rem(12px);
-            margin: 0 0 0 rem(-6px);
+            // Account for border width
+            margin: -1px 0 -1px calc(#{rem(-6px)} -1px);
```
cassidyjames, line 146: Nitpick: if it's subtraction and not passing `-1px` in, there should be spaces around the `-`.

```suggestion
            margin: -1px 0 -1px calc(#{rem(-6px)} - 1px);
```

src/widgets/_scales.scss
```diff
@@ -131,13 +142,15 @@ scale {
...
         }
 
         highlight.bottom,
         fill.bottom {
             border-radius: 0 rem(12px) rem(12px) 0;
-            margin: 0 rem(-6px) 0 0;
+            // Account for border width
+            margin: -1px calc(#{rem(-6px)} -1px) -1px 0;
```
cassidyjames, line 153: Subtraction nitpick

```suggestion
            margin: -1px calc(#{rem(-6px)} - 1px) -1px 0;
```

src/widgets/_scales.scss
```diff
@@ -160,13 +173,15 @@ scale {
         highlight.top,
         fill.top {
             border-radius: rem(12px) rem(12px) 0 0;
-            margin: rem(-6px) 0 0;
+            // Account for border width
+            margin: calc(#{rem(-6px)} -1px) -1px 0 -1px;
```
cassidyjames, line 177: Subtraction nitpick

```suggestion
            margin: calc(#{rem(-6px)} - 1px) -1px 0 -1px;
```

src/widgets/_scales.scss
```diff
@@ -160,13 +173,15 @@ scale {
...
         }
 
         highlight.bottom,
         fill.bottom {
             border-radius: 0 0 rem(12px) rem(12px);
-            margin: 0 0 rem(-6px);
+            // Account for border width
+            margin: 0 -1px calc(#{rem(-6px)} -1px) -1px;
```
cassidyjames, line 184: Subtraction nitpick

```suggestion
            margin: 0 -1px calc(#{rem(-6px)} - 1px) -1px;
```

src/_index.scss
danrabbit: Yeah, you're smart and I have no idea what I was thinking lmao

cassidyjames (changes requested):
Minor nits

cassidyjames (changes requested):
I can also confirm @hanaral's issue.

Screenshot from 2020-10-12 16-19-29@2x

cassidyjames (approved):
Appears to do the thing

PR #842, opened 2020-10-03
Avatars: Don't draw background, neutral border with images

danrabbit:
Screenshot from 2020-10-02 17 10 45@2x

### DIFF
--- a/src/widgets/_avatars.scss
+++ b/src/widgets/_avatars.scss
@@ -14,22 +14,17 @@ avatar {
     border-color: #{'alpha(@accent_color_500, 0.5)'};
     color: #{'@accent_color_900'};
 
-    &.color1 {
-        background-color: $STRAWBERRY_300;
-        border-color: rgba($STRAWBERRY_500, 0.5);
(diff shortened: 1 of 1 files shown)


Exalm:
Unrelated, but you only have 8 colors? `HdyAvatar` has 14, you need to specify all of them unless you want the remaining style classes to be unstyled. If you only want to use 8 colors, add some duplicates, like color9 == color1 etc.

Exalm:
Screenshot from 2020-10-03 17-57-32

Here are names to reproduce all 14.

fabiozaramella:
I can confirm that Hdy.Avatar widgets now looks as it should with this PR. However, 2 colors are missing/duplicate when tested using the above-mentioned names (see GIF attachment). Except this, everything is perfect. 👍🏼️

Peek 08-10-2020 10-41

fabiozaramella (approved):

fabiozaramella:
I thinks that the new light blue and yellow/orange work really well 😍️. Good job!

Issue #827, opened 2020-09-15
Transparent/broken labels in AppCenter headings


cassidyjames:
So this was part of #636 I think, but I'm seeing it again since I think we reverted the specific label styling. Headings in AppCenter app listings appear transparent until the window changes focus.

cassidyjames:
![Screenshot from 2020-09-15 12-45-59](https://user-images.githubusercontent.com/611168/93251442-6b21a480-f751-11ea-82af-e524c3037c7a.png)

danrabbit:
Unsure if related but there are also issues with the labels in epiphany's auto complete drop down

danrabbit:
I can't reproduce this anymore. I think we fixed it

PR #844, opened 2020-10-05
Windows: Style .background

danrabbit:
Fixes some busted styles: GtkAssistant, background in Tasks, probably more


cassidyjames:
Ooh, this looks like it fixes the transparent text in Epiphany mentioned in #827… but not in AppCenter.

cassidyjames (approved):

PR #847, opened 2020-10-06
Support flat entry styles

danrabbit:
For Tasks. Probably not the most optimal flat entry style ever, but gets the ball rolling


cassidyjames (approved):
Does the thing

Issue #734, opened 2020-06-06
greenfield: Should export the same public colors as Adwaita


Exalm:
The full list of those colors, minus the WM ones and content_view_bg:

- [x] theme_fg_color
- [x] theme_text_color
- [x] theme_bg_color
- [x] theme_base_color
- [x] theme_selected_bg_color
- [x] theme_selected_fg_color
- [x] insensitive_bg_color
- [x] insensitive_fg_color
- [x] insensitive_base_color
- [x] borders
- [x] theme_unfocused_fg_color
- [x] theme_unfocused_text_color
- [x] theme_unfocused_bg_color
- [x] theme_unfocused_base_color
- [ ] theme_unfocused_selected_bg_color
- [ ] theme_unfocused_selected_fg_color
- [x] unfocused_insensitive_color
- [x] unfocused_borders
- [x] warning_color
- [x] error_color
- [x] success_color

While there isn't any API, these colors are used as a sorta-kinda-API, and lack of them causes serious breakage in many places. For example WebKit's back gesture uses `@theme_base_color` for the fallback color when there's no screenshot. It does fall back to white, but the main reason it uses that color is for dark mode support :)

PR #848, opened 2020-10-07
Introduce and use insensitive-fg-color

### LINKED ISSUE #689: Insensitive image buttons in headerbars are not distinct enough
i.e. back/forward buttons in Epiphany, or the BrowserButton in Ephemeral. For image buttons, I think it makes sense to drop the opacity (which is what I do in my custom styles for Ephemeral) to distinguish it.

danrabbit:
Fixes #689 

One more for #734


cassidyjames (approved):
Dope

PR #851, opened 2020-10-07
Buttons: Support button.color

### LINKED ISSUE #806: Style color buttons
Screenshot from 2020-09-08 18 28 58@2x

danrabbit:
Fixes #806

### DIFF
--- a/src/widgets/_buttons.scss
+++ b/src/widgets/_buttons.scss
@@ -106,10 +106,18 @@ button {
         color: #{'@placeholder_text_color'};
     }
 
+    &.color,
     &.combo,
     &.image-button,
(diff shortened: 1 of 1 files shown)


cassidyjames:
Is this button supposed to match the size of the combobox? Or what's being fixed? 

I feel like something weird is still happening here. I don't see a difference at first, but if I mess with font size, the button changes size—but not in a totally predictable way. Maybe GTK itself is doing something weird size-wise?

danrabbit:
It just makes the padding the same on all sides and makes the little color area rounded

cassidyjames (approved):
Ah, I see it now. Dope.

PR #863, opened 2020-10-21
Show selection and focus in Grids and Lists

### LINKED ISSUE #781: Focus/selection styles in treeview
We have focus/selection styles for Gtk.ListBox but not Gtk.Treeview

danrabbit:
Fixes #781 

Yo dawg I heard you like to get around with a keyboard

Shows focus and selection in flowboxes, listboxes, icon views, and treeviews such as in:
* Files
* System Settings
* Music
* Videos
* And More!

Does not work for the applications menu because that is a bunch of buttons in a grid. Gross

### DIFF
--- a/src/widgets/_lists.scss
+++ b/src/widgets/_lists.scss
@@ -6,13 +6,13 @@ list {
     }
 
     row {
+        &:focus,
         &:selected {
-            background-color: rgba($fg-color, 0.1);
+            background: rgba($fg-color, 0.15);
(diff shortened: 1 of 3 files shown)


cassidyjames (approved):
For some reason my brain is telling me that focus and selected should not always necessarily be the same… maybe in the case of multi-select or something? But for now, this is a huge usability improvement and we can address that if/when it comes up.

Issue #692, opened 2020-05-14
New stylesheet headerbar back button lacks arrow 


hanaral:
Old stylesheet:
![image](https://user-images.githubusercontent.com/58219504/81948463-ad977100-95f9-11ea-80b2-f42a05a4aec5.png)
New stylesheet:
![image](https://user-images.githubusercontent.com/58219504/81948570-ca33a900-95f9-11ea-82c0-6ac214f9eb7c.png)

cassidyjames:
Yep, as expected at this stage. The arrow style we did before has to be provided as an SVG that perfectly matches the final GtkCSS styling, and since the latter is in flux, it doesn't make sense to try and keep up with a hand-crafted SVG. Ideally we could use Granite to actually custom draw a back button, but I'm not sure if that's the concrete plan or not.

hanaral:
Alright, I assumed the SVG wasn't done yet (I will admit I *did* fear that it might be going away)

PR #864, opened 2020-10-23
Buttons: Add back button support

danrabbit:
Not exactly the fancy shaped buttons, but works with dark/light and will work with colors etc

### DIFF
--- a/src/widgets/_buttons.scss
+++ b/src/widgets/_buttons.scss
@@ -104,6 +104,39 @@ button {
         }
     }
 
+    &.back-button {
+        background-repeat: no-repeat no-repeat;
+        background-size: 16px, cover;
+
+        &:dir(ltr) {
+            background-image:
+                -gtk-icontheme('go-previous-symbolic'),
+                linear-gradient(
+                    to bottom,
+                    #{'alpha(@highlight_color, 0.2)'},
+                    rgba(white, 0)
+                );
+            padding-left: calc(#{rem(9px)} + 16px);
+            background-position:
+                #{rem(6px)} 50%,
+                center, center;
+        }
+
+        &:dir(rtl) {
+            background-image:
+                -gtk-icontheme('go-next-symbolic'),
+                linear-gradient(
+                    to bottom,
+                    #{'alpha(@highlight_color, 0.2)'},
+                    rgba(white, 0)
+                );
+            padding-right: calc(#{rem(9px)} + 16px);
+            background-position:
+                calc(100% - #{rem(6px)}) 50%,
+                center, center;
+        }
+    }
+
     &:disabled {
         color: $insensitive-fg-color;
     }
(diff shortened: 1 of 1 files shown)


danrabbit:
Actually maybe this is evil. If we want to go this route it would probably be better to properly add an image to the button

cassidyjames:
This looks like what we did in the Pop! stylesheet to support Granite back buttons, and look like it's what they're still doing. I am fine with it as a stopgap and/or if we want to be able to more easily change the style later. But if we're sure this is what we want to do, then yeah, maybe we just have a `Granite.BackButton` or something that we use instead that adds the icon… or just have apps add the icon themselves by convention.

I personally really like the shaped icon for the style class, but I know it's hard to do, especially now supporting font sizes and accent colors and everything. If we can't do it programatically/without making a ton of images, than packing a `←` icon in there is fine with me.

cassidyjames (approved):
Tentatively approving because this supports the existing style class.

danrabbit:
Yeah what I haven't done is look into what it would take to make a Granite.BackButton with custom drawing. That would be the way to really make sure we can work correctly with colors and everything

hanaral:
I understand that it's a chore to get the custom shape to not just be an SVG, but there is a severe loss in design paradigm if they are dropped completely. I think that there should be some more dialogue on how it can be achieved, would @micahilbery know anything about CSS black magic I wonder

micahilbery:
@hanaral Unfortunately, I don't think this is possible with gtk-css alone. Gtk-css doesn't support the entirety of normal css and therefore doesn't support things like pseudo-elements that would make this possible. You'd need a custom drawing/widget which is unfortunately way over my head.

hanaral:
> @hanaral Unfortunately, I don't think this is possible with gtk-css alone. Gtk-css doesn't support the entirety of normal css and therefore doesn't support things like pseudo-elements that would make this possible. You'd need a custom drawing/widget which is unfortunately way over my head.
Ah that sucks, cheers for clearing that up. I imagine it is not impossible to have a specific back button widget made for Granite.

src/widgets/_buttons.scss
```diff
@@ -104,6 +104,39 @@ button {
         }
     }
 
+    &.back-button {
```
cassidyjames, line 107: Is it dumb to leave a reminder here?

```suggestion
    // Stopgap since we can't do angled buttons in GtkCSS. Ideally we would 
    // deprecate this in favor of something like a Granite.BackButton with
    // custom drawing. Until then, stick a back icon into .back-button.
    &.back-button {
```

cassidyjames:
@danrabbit do we consider this a fix for #692, or do we want to keep it open for a better solution?

danrabbit:
@cassidyjames I havent really made it a priority to look into a custom drawing solution for back arrows. I would prefer not to try to do images with all the accent variants etc. So I guess I'd like to keep it open. Maybe move it to Granite?

cassidyjames:
@danrabbit yeah that sounds like a better place to track the underlying issue.

PR #865, opened 2020-10-23
Typography: Allow specifying accent color

danrabbit:
This will probably be the best fix for https://github.com/elementary/sideload/issues/98

Otherwise we have to load stylesheets in and out or something complicated

Once this is merged we can get rid of the custom stylesheet in sideload and just set `.accent.orange` and `.accent.purple` etc

Screenshot from 2020-10-23 14 27 20


cassidyjames (approved):
:+1:

PR #868, opened 2020-10-25
Fix paned separator

msmaldi:
Currently the course only appears on the separator and 8 pixels after the separator.
The correction makes the cursor appear 4 pixels before and 4 pixels after.

This fix issue [573](https://github.com/elementary/stylesheet/issues/573)


danrabbit:
Thanks for your branch. This is intentional to make sure we don't interfere with scroll bars

PR #869, opened 2020-10-26
Windows: Remove box shadow on .unified .flat.titlebar

danrabbit:
Fixes a funny extra dark shadow line in the separators in Tasks (and other apps)

We don't need any box shadow at all here because the window border is set on the unified window and the inset highlight is done on the overlay node


cassidyjames:
@danrabbit can you share a screenshot pointing out the difference here? I'm not seeing it. :mag:

danrabbit:
Screenshot from 2020-11-03 09 32 56@2x

danrabbit:
@cassidyjames actually this doesn't fix that thing. I don't know why I thought it did

cassidyjames (changes requested):
Doesn't do the thing

PR #875, opened 2020-11-03
Menus: make switches a wee bit smaller

danrabbit:


cassidyjames (commented):
They're cute, I like them. Looks better in apps like Ephemeral, too.

cassidyjames (approved):

PR #878, opened 2020-11-04
Style Gtk.Calendar

danrabbit:
This widget sucks and there's not much we can actually do here


cassidyjames (approved):
Looks a lot better than what we have!

PR #880, opened 2020-11-10
Buttons: scale image buttons when pressed

danrabbit:
because whimsy!

### DIFF
--- a/src/_animate.scss
+++ b/src/_animate.scss
@@ -29,3 +29,15 @@ entry.spin:dir(rtl) image:first-child {
         -gtk-icon-transform: rotate(1turn);
     }
 }
+
+@keyframes press {
+    0% {
+        -gtk-icon-transform: scale(1);
+    }
+    50% {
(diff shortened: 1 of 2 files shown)


cassidyjames (changes requested):
This really only seems to work if you press-and-hold on buttons, at least on my hardware. Otherwise, it seems to start to animate but then bails when you stop pressing. Not sure if that's a GTK thing or just a CSS animation thing.

cassidyjames:
@danrabbit okay this actually seems to work really well in the Applications Menu and I love it there; I wonder if we do this in its own internal styling to start, instead of on all imagebuttons?

hanaral:
It looks really nice with symbolic buttons as they look like physical objects being depressed - but I'm not sure if it looks so great with colour buttons. 
My impression is that colour buttons are a bit like context menu buttons where they don't react to being clicked (although I did open an issue for something that could solve this in #950 like macOS does)

PR #888, opened 2020-11-13
Entries: Set minimum height to accomodate icons

danrabbit:
This prevents a resize jump when adding icons to entries such as in User Accounts when creating a new user

### DIFF
--- a/src/widgets/_entries.scss
+++ b/src/widgets/_entries.scss
@@ -26,6 +26,8 @@
 
 entry {
     @extend %entry;
+    // Prevent size change when setting primary or secondary icons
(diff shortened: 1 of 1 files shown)


cassidyjames (approved):
This works but has implications for small text; it means the entries never get smaller. Would it be evil to scale icons in entries by rems so it also gets smaller there, and then use rems here?

danrabbit:
@cassidyjames I think like we said before the problem is we can only scale the texture, but we can't scale the size of the actual Gtk.Image. I'm pretty sure it's hardcoded at 16px

PR #894, opened 2020-12-03
Spinners: spin discretely

cassidyjames:
Our spinner icon is designed for a discrete 12-frame spin, but we're currently smoothly animating the whole spin. This PR sets the animation-timing-function for spinners specifically to take 12 steps, also matching the new BGRT plymouth theme.

This is hard to capture in a GIF because of framerate, but here's an attempt; notice in the "after" GIF the spokes stay in the correct orientation.

Before | After
---|---
ezgif com-video-to-gif | ezgif com-video-to-gif(1)

### DIFF
--- a/src/widgets/_spinners.scss
+++ b/src/widgets/_spinners.scss
@@ -5,7 +5,7 @@ spinner {
 
     &:checked {
         opacity: 0.8;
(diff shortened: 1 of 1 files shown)


danrabbit (approved):
You're so smart :heart:

PR #897, opened 2020-12-07
Fix editable treeview cell background color

### LINKED ISSUE #896: Renaming items in sidebars shows transparent entry on top of existing item
I'm not sure if this is a stylesheet issue, but I think so: if I try to rename an item in a sidebar (i.e. in Files or Code), I get a transparent (and thus unreadable) entry on top of the existing item in the sidebar.

danrabbit:
Fixes #896 

We can't use transparent backgrounds here in treeview because I guess the widgets are stacked

### DIFF
--- a/src/widgets/_sidebars.scss
+++ b/src/widgets/_sidebars.scss
@@ -40,4 +40,9 @@ treeview.sidebar {
     .view {
         @extend %sidebar-row;
     }
+
+    entry.flat {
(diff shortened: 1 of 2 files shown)


cassidyjames (approved):
Yay, does the thing! :tada: I can rename in Code again :stuck_out_tongue_closed_eyes:

PR #899, opened 2020-12-09
Dim backdrop buttons

cassidyjames:
This might not be the best approach, but currently, headerbars only change their color a bit when in backdrop, and accent-colored widgets drop their accent color. However, this leaves apps with headerbars and content in a more ambigous state.

This PR dims buttons in the backdrop state to make it even more clear which app is focused. There are probably lots of other small backdrop improvements we could do, but this might be a good step.

Feedback welcome.

### DIFF
--- a/src/widgets/_buttons.scss
+++ b/src/widgets/_buttons.scss
@@ -145,6 +145,10 @@ button {
         color: $insensitive-fg-color;
     }
 
+    :backdrop & {
+        opacity: 0.75;
+    }
+
     &.color,
     &.combo,
     &.image-button,


danrabbit (changes requested):
Yeah, so this isn't the best way to do this. Labels and images may also have their own backdrop values, so dropping the button opacity would compound that effect. Buttons also extend `outset-background` which has values for backdrop states already

A better approach would be to add `-gtk-icon-effect: dim` here so that we don't compound opacity effects and to adjust the values of `outset-background` to be more dramatically different

PR #900, opened 2020-12-10
Make backdrop accent images text colored

danrabbit:
I forgot to uncomment some code and somebody approved it anyways!!

Also, maybe we want to do this for all backdrop images instead of #899 ?

### DIFF
--- a/src/_typography.scss
+++ b/src/_typography.scss
@@ -121,10 +121,10 @@ label.h4,
         }
     }
 
-    // &:backdrop {
-    //     color: $fg-color;
-    //     opacity: 0.6;
-    // }
+    &:backdrop {
+        color: $fg-color;
+        opacity: 0.6;
+    }
 }
 
 .dim-label,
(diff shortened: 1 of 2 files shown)


cassidyjames:
Oops. Yeah this could make sense. But the other branch affects things like headerbar stack switchers and raised buttons, which does still improve that distinction between focused and unfocused. But I am open to anything we want to try.

danrabbit:
Ah okay yeah. For the buttons themselves they should already have a backdrop gradient so that should be adjusted in those styles

I think we don't currently do anything with backdrop labels but we should probably apply that globally as well and not just to buttons

hanaral:
Not sure if it is this branch's code, but the volume and brightness overlays look borked:
Screenshot from 2020-12-16 09-53-02
2020-12-16 09 54 42 のスクリーンショット

cassidyjames (approved):

PR #902, opened 2020-12-14
Focus styles

### LINKED ISSUE #730: Inputs and buttons lack focus style
The coloured outline and inset glow that input/selection fields and buttons had in the old stylesheet is not present in the new one.

danrabbit:
Fixes #730 

Introduces focus style for:
* Buttons
* Checkboxes
* Radios
* Entries
* Spinbuttons

Does not include
* Scales
* Switches
* Modelbuttons
* probably lots of other stuff

### DIFF
--- a/src/widgets/_buttons.scss
+++ b/src/widgets/_buttons.scss
@@ -1,6 +1,7 @@
 button {
     background-clip: padding-box;
     border-radius: rem(3px);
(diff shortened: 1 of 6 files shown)


cassidyjames (approved):
Yay! This is a huge improvement to not having any. I am seeing some cases we might not be handling like linked buttons, but this is already looking really good. Nice work.

PR #919, opened 2021-01-01
Add: HdyViewSwitcher styling

### LINKED ISSUE #917: Hdy.ViewSwitcher: Completely broken styling
## What Happened

The widget HdyViewSwitcher in libhandy looks to be completely unstyled, leading to breakage in apps that use it and aren't Adwaita themed.

old

## Expected Behavior

It should look just like it does in Adwaita, where it is full-height and linked, maybe with some elementary OS touch.

expect

## Steps to Reproduce

1. Open an app that does not use Adwaita as theme, but uses libhandy HdyViewSwitcher
2. Look at how not-styled it is
3. ???
4. Profit

## Platform Information

elementary OS 6.0 Early Access

lainsce:
This adds some starting styling for the widget HdyViewSwitcher.

Fixes #917

### DIFF
--- a/src/widgets/_viewswitcher.scss
+++ b/src/widgets/_viewswitcher.scss
@@ -74,9 +74,7 @@ viewswitcher {
     }
 }
 
-headerbar {
-    viewswitcher {
-        button {
+headerbar viewswitcher button {
             &:not(:checked):hover:not(:backdrop) {
                 background-image: linear-gradient(
                     to top,
--- a/src/widgets/_viewswitcher.scss
+++ b/src/widgets/_viewswitcher.scss
@@ -1,5 +1,4 @@
-viewswitcher {
-    & button {
+viewswitcher button {
         margin: 0;
         padding: 0;
         border-radius: 0;
--- a/src/widgets/_viewswitcher.scss
+++ b/src/widgets/_viewswitcher.scss
@@ -6,7 +6,7 @@ viewswitcher button {
     border-bottom: 0;
     box-shadow: none;
     font-size: rem(13px);
-    -gtk-icon-palette: error #{'@text_color'};, success #{'@text_color'};, warning #{'@text_color'};;
+    -gtk-icon-palette: error @text_color, success @text_color, warning @text_color;
 
     &:not(:checked):not(:hover) {
         background: transparent;
--- a/src/widgets/_viewswitcher.scss
+++ b/src/widgets/_viewswitcher.scss
@@ -1,0 +1,103 @@
+viewswitcher {
+    & button {
+        margin: 0;
+        padding: 0;
+        border-radius: 0;
+        border-top: 0;
+        border-bottom: 0;
+        box-shadow: none;
+        font-size: rem(16px);
+
+        &:not(:checked):not(:hover) {
+            background: transparent;
+        }
+
+        &:not(:only-child):not(:last-child) {
+            border-right-width: 0;
+        }
+
+        &:not(only-child):first-child:not(:checked):not(:hover),
+        &:not(:checked):not(:hover) + &:not(:checked):not(:hover) {
+            border-left-color: transparent;
+        }
+
+        &:not(only-child):last-child:not(:checked):not(:hover) {
+            border-right-color: transparent;
+        }
+
+        &:not(:checked):hover:not(:backdrop) {
+            background-image: #{'image(lighter(@color_primary))'};
+        }
+
+        &:not(only-child):first-child:not(:checked):hover,
+        &:not(:checked):hover + &:not(:checked):not(:hover),
+        &:not(:checked):not(:hover) + &:not(:checked):hover {
+            border-left-color: #{'shade(@color_primary, 1.15)'};
+        }
+
+        &:not(only-child):last-child:not(:checked):hover {
+            border-right-color: #{'shade(@color_primary, 1.15)'};
+        }
+
+        &:not(:checked):hover:backdrop {
+            background-image: #{'image(@color_primary)'};
+        }
+
+        stack > box {
+
+            & label.active {
+                font-weight: bold;
+            }
+
+            &.narrow {
+                font-size: rem(12px);
+                padding-top: rem(7px);
+                padding-bottom: rem(5px);
+
+                & image,
+                & label {
+                    padding-left: rem(8px);
+                    padding-right: rem(8px);
+                }
+            }
+
+            &.wide {
+                padding: rem(8px) rem(12px);
+
+                & image,
+                & label {
+                    padding-left: rem(7px);
+                    padding-right: rem(7px);
+                }
+            }
+        }
+    }
+}
+
+headerbar {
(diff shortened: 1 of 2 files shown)


src/widgets/_viewswitcher.scss
```diff
@@ -0,0 +1,103 @@
...
+                    padding-right: rem(7px);
+                }
+            }
+        }
+    }
+}
+
+headerbar {
+    viewswitcher {
+        button {
```
danrabbit, line 79: It's not necessary to use nesting like this if there's only one child. You could save a whole lot of indentation with
```suggestion
headerbar viewswitcher button {
```

src/widgets/_viewswitcher.scss
```diff
@@ -0,0 +1,103 @@
...
+        stack > box {
+
+            & label.active {
+                font-weight: bold;
+            }
+
+            &.narrow {
+                font-size: rem(12px);
+                padding-top: rem(7px);
+                padding-bottom: rem(5px);
```
danrabbit, line 55: Why is there more padding on top?

src/widgets/_viewswitcher.scss
```diff
@@ -0,0 +1,103 @@
+viewswitcher {
+    & button {
```
danrabbit, line 2: This could just be 
```suggestion
viewswitcher button {
```

src/widgets/_viewswitcher.scss
```diff
@@ -0,0 +1,103 @@
...
+        margin: 0;
+        padding: 0;
+        border-radius: 0;
+        border-top: 0;
+        border-bottom: 0;
+        box-shadow: none;
+        font-size: rem(16px);
```
danrabbit, line 9: Why are you making the text much larger here? That's like really huge

src/widgets/_viewswitcher.scss
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        stack > box {
+
+            & label.active {
+                font-weight: bold;
+            }
+
+            &.narrow {
+                font-size: rem(12px);
```
danrabbit, line 53: This seems like it should probably instead be
```suggestion
                @extend .small-label
```

src/widgets/_viewswitcher.scss
```diff
@@ -0,0 +1,103 @@
...
+                & label {
+                    padding-left: rem(8px);
+                    padding-right: rem(8px);
+                }
+            }
+
+            &.wide {
+                padding: rem(8px) rem(12px);
+
+                & image,
```
danrabbit, line 67: There is a LOT of padding on images here:
Screenshot from 2021-01-05 13 31 18@2x

typically images should have 3px of padding

danrabbit (changes requested):
Have you tested these styles? The text sizes are really huge and it doesn't look that good in HdyDemo

lainsce:
> Have you tested these styles? The text sizes are really huge and it doesn't look that good in HdyDemo

I did test them under LoDPI, maybe the text being huge and image padding being weird is because of that. I'll reduce text size and fix the image padding.

lainsce:
It now should look less weird. Tested on HiDPI Handy Demo. :)

lainsce:
Also, what's with the @extend error? Should I import something to make it work? I'm new at Sass :P

src/widgets/_viewswitcher.scss
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+}
+
+headerbar viewswitcher button {
+    &:not(:checked):hover:not(:backdrop) {
+        background-image: linear-gradient(
+            to top,
+            #{'shade(alpha(@color_primary, 0.7), 0.96)'} rem(2px),
+            #{'alpha(@color_primary, 0.7))'}
```
danrabbit, line 80: Extra `)`
```suggestion
            #{'alpha(@color_primary, 0.7)'}
```

danrabbit (changes requested):
Looking at the demo, I think we should probably reset `-gtk-icon-pallete` here. It doesn't seem to make sense to have icon colors. So a line like:

```
-gtk-icon-palette: error @text_color, success @text_color, warning @text_color;
```

src/widgets/_viewswitcher.scss
```diff
@@ -0,0 +1,97 @@
+viewswitcher button {
+    margin: 0;
+    padding: 0;
+    border-radius: 0;
+    border-top: 0;
+    border-bottom: 0;
+    box-shadow: none;
+    font-size: rem(10px);
```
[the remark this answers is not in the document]
lainsce: It's to accomodate large titles for example "Interface" or "Stopwatch". They'd get very cramped otherwise.

src/widgets/_viewswitcher.scss
```diff
@@ -0,0 +1,97 @@
...
+
+        &.wide {
+            padding: rem(3px) rem(3px);
+
+            & image,
+            & label {
+                padding-left: rem(7px);
+                padding-right: rem(7px);
+            }
+        }
```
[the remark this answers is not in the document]
lainsce: Alrighty

src/widgets/_viewswitcher.scss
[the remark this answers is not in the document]
lainsce: Also, I did an oops, it should be 13px.

src/widgets/_viewswitcher.scss
```diff
@@ -0,0 +1,92 @@
...
+
+        &.wide {
+            padding: 0 9px;
+        }
+    }
+}
+
+headerbar viewswitcher button {
```
danrabbit, line 70: You can actually nest this inside the block above as
```
headerbar & {
```

the `&` represent the parent selector

src/widgets/_viewswitcher.scss
```diff
@@ -0,0 +1,92 @@
+viewswitcher button {
+    margin: 0;
+    padding: 0;
+    border-radius: 0;
+    border-top: 0;
+    border-bottom: 0;
+    box-shadow: none;
+    font-size: rem(13px);
```
danrabbit, line 8: I am once again asking why you are setting a different font size here at all?

lainsce: Try it in the demo, make one of the viewswitcher titles a really long one, and test with and without.
That should explain why.

lainsce:
Once this is merged, a follow-up should occur in regards to the headerbar, to use these:

```css
    headerbar {
	padding: 0 6px;
    }

    headerbar button:not(.radio) {
        margin: 6px;
        padding: 6px
    }

    headerbar entry.search {
	margin: 6px 0;
    }
```

src/widgets/_viewswitcher.scss
lainsce: After some testing, I think I'll drop this, might be fine without this.

danrabbit (approved):
Better than master :+1:

PR #922, opened 2021-01-04
Buttons: Remove link padding

### LINKED ISSUE #921: Link buttons should not get padding
## What Happened

When I use a LinkButton in a layout, it gets button padding even though it's not styled as a button. This makes it difficult to align with other text-based elements.

Screenshot from 2021-01-04 14 53 52

## Expected Behavior

Link buttons don't get any extra padding, more like a label than a button.

Screenshot from 2021-01-04 15-04-34

```css
.text-button.link {
    padding: 0;
}
```

## Platform Information

elementary OS 6 daily build

danrabbit:
Fixes #921

### DIFF
--- a/src/widgets/_buttons.scss
+++ b/src/widgets/_buttons.scss
@@ -171,6 +171,10 @@ button {
         padding: rem(4px);
     }
 
+    &.text-button.link {
+        padding: 0;
+    }
+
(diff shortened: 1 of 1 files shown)


cassidyjames (approved):
Thank you! :)

PR #923, opened 2021-01-04
Buttons: Add arrows to links

danrabbit:
Screenshot from 2021-01-04 15 40 00@2x

Might be kind of extra, but I think it's cute

### DIFF
--- a/src/widgets/_buttons.scss
+++ b/src/widgets/_buttons.scss
@@ -172,7 +172,28 @@ button {
     }
 
     &.text-button.link {
+        background-image: -gtk-icontheme('go-next-symbolic');
+        background-repeat: no-repeat;
+        background-size: 16px;
         padding: 0;
+
+        &:dir(ltr) {
+            padding-right: calc(#{rem(9px)} + 16px);
+            background-position: calc(100% - #{rem(6px)}) 50%;
+
+            &:hover {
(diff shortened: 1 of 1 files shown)


cassidyjames (changes requested):
I _think_ we only want to do this for links where we explicitly opt into it, like with an `.external` class or something, right? I feel like this would get pretty wild when used elsewhere, especially when linking to things within the same app, like with Switchboard Plugs.

megatux:
For me, these arrows looks a bit like *"next"* action.
A quick img search on the web for "external link" seems to show a box with an arrow pointing to top-right corner as the most popular representation, followed by a two-links-chain (like the one in this Github comment toolbar & emoji  :link:)

240_F_393233529_8Q44R3IrfTrpnci7Vu5kqwHXj2UDWTWU
240_F_231486280_NCwG2QK6shff6CHUBJapZKQIArk5Y41l

What do you think of using one of these icon representations?

hanaral:
> 240_F_231486280_NCwG2QK6shff6CHUBJapZKQIArk5Y41l
> 
> What do you think of using one of these icon representations?

This one looks like an 'Export' button, or open in button, which isn't quite an indication of a URL or link.

danirabbit:
Yeah, I might revisit this since we have an external link icon now in Granite

PR #927, opened 2021-01-12
Menus: Fix menu scroll arrows

danrabbit:
This doesn't fix the issue with small menus in the applications menu, but it does fix showing the scroll arrows


cassidyjames (approved):
Nice

PR #928, opened 2021-01-13
Experiment: Icon Dropshadow CSS class

lainsce:
The class `.icon-dropshadow` adds the same shadow as the one made manually following the guidelines,
but automatically, so nobody would need to make the shadow themselves in Inkscape anymore.

The catch, is that one would need to add the class targetting the `Gtk.Image` holding the icons.

Looks like this:
result

P.S. If the icons above look bad in their shadow, it's due to the manually added shadow plus the css class' shadow
on top of each other. Check the Icon Previewer icon for the intended effect.

### DIFF
--- a/src/widgets/_index.scss
+++ b/src/widgets/_index.scss
@@ -35,9 +35,9 @@ selection {
 .icon-dropshadow {
     // Used for adding shadow to icons programmatically and not manually,
     // mimics the most used shadow type in icons.
-    -gtk-icon-shadow: 0 4px 3px rgba(black, 0.0),
-                      0 4px 2px rgba(black, 0.12),
-                      0 4px 1px rgba(black, 0.0);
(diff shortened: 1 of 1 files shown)


pongloongyeat:
Have the icons always looked this popped out? Or has it always been like this cause this looks amazing

lainsce:
> Have the icons always looked this popped out? Or has it always been like this cause this looks amazing

No, the "popped effect" is due to using this css class. I still need to change some parameters for those shadows it seems, because it shouldn't pop-out, rather it should look as if there's a light cast from above.

lainsce:
Screenshot from 2021-01-13 11 17 59

Fixed 'em.

lainsce:
No matter what I do I can't find a way to have the intended dropshadow for circle icons. Closing the experiment.

PR #936, opened 2021-01-25
progressbars: animate horizontal progress

cassidyjames:
Fixes https://github.com/elementary/stylesheet/discussions/935. Only animates horizontal progressbars because it looked weird with vertical ones no matter which direction I did—and I wasn't sure we really use them in that sort of context anyway.

The idea behind the animation is to start a glow in the middle, then shoot it to the end of the progressbar, helping emphasize forward progression.

Might need RTL work?

### DIFF
--- a/src/widgets/_progressbars.scss
+++ b/src/widgets/_progressbars.scss
@@ -38,8 +38,13 @@ progressbar {
         }
 
         progress {
+            animation: progress 2s ease infinite;
             margin-left: -1px;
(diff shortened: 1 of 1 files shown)


danrabbit:
How about https://github.com/elementary/stylesheet/pull/937

danrabbit:
Gonna close since #937 is a bit more fleshed out and we can iterate there

PR #937, opened 2021-01-26
Progressbars: animate .horizontal:not(.pulse)

danrabbit:


src/widgets/_progressbars.scss
```diff
@@ -32,14 +32,43 @@ progressbar {
...
+                        rgba(white, 0.25) 60%,
+                        rgba(white, 0)
+                    );
+
+                &:dir(ltr) {
+                    animation: progress 2s ease infinite;
+                }
+
+                &:dir(rtl) {
+                    animation: progress-rtl 2s ease infinite;
```
btkostner, line 64: I'm not sure if it's supported, but you might save yourself a bit of code by using `animation-direction: reverse`

https://developer.mozilla.org/en-US/docs/Web/CSS/animation-direction

danrabbit: You so smart

hanaral:
Don't wanna be that guy, but lightly shaded diagonal stripes like the old macOS progressbar would be pretty cute:
Screenshot from 2021-01-26 18-18-17

src/widgets/_progressbars.scss
```diff
@@ -32,14 +32,40 @@ progressbar {
...
         trough {
             @extend %outset-background;
         }
 
         progress {
             margin-left: -1px;
             margin-right: -1px;
+
+            &:not(.pulse) {
+                animation: progress 2s ease infinite;
```
cassidyjames, line 44: After playing with a bunch of values, this feels nicer to me. As a bonus, it's the same easing used elsewhere.

```suggestion
                animation: progress 1.5s $easing infinite;
```

src/widgets/_progressbars.scss
```diff
@@ -32,14 +32,40 @@ progressbar {
...
+                        rgba(white, 0)
+                    ),
+                    linear-gradient(
+                        to right,
+                        rgba(white, 0),
+                        rgba(white, 0.25) 60%,
+                        rgba(white, 0)
+                    );
+
+                &:dir(rtl) {
```
cassidyjames, line 60: It turns out GTK automatically handles the animation direction for RTL… but not for the inverted-style progress bars… :facepalm: So this should be:

```suggestion
                &.right {
```

cassidyjames (changes requested):
I like this a lot. Couple of tiny tweaks, then it's good imho.

cassidyjames (approved):
hawt

Issue #903, opened 2020-12-14
Flash with focus outline in dark style


cassidyjames:
When a button is focused in the dark style, it looks like it flashes light briefly before showing the accent color outline.

okitoxo:
This was proofed on Odin? I can't replicate on Hera, building the stylesheet and using Gtk Inspector.

okitoxo:
The problem
This widget have a trasition for all properties and it doesn't have a initial value of outline.

We have two ways:

- Mantain transition for all properties (other widgets don't have that behaivor in focus outlined) and set initial values for outline property.
- Restrict transition for only backgroud property (for active state)

I think second way is right to keep uniform widgets.

Issue #749, opened 2020-06-17
Destructive action button flahses when pressed in dark style


cassidyjames:
I noticed in #741 but couldn't figure out what was causing it. I'm seeing a fast flash like between two states when clicking the destuctive action button in the dark style. It might also be in the light style but harder to spot. I'm wondering if it has something to do with transitions between states, and Gtk not executing those at the exact same time?

okitoxo:
Is necesary apply transition for outline buttons when are focused?
This issue happens because button don't have an initial value for outline, animation on transition not work fine.

PR #958, opened 2021-02-27
set transition for only background

okitoxo:
Restrict transition for only backgroud property of button (for active state). #903


okitoxo:
This change fix #903 and #749

cassidyjames:
I'll have to look at this more closely, but my instinct says we'd also want to transition and shadows and borders. I agree that a more restrictive scope is probably better than `all`, though—it makes us actually consider what gets transitioned instead of treating it like a wildcard.

cassidyjames (approved):

PR #965, opened 2021-03-09
Sidebar Dark: Set bg_color(2) on selected item

### LINKED ISSUE #822: Sidebar selection not visible in dark style
Screenshot from 2020-09-11 12 50 41@2x

Dunno what happened there

okitoxo:
Selected item on sidebar dark style not visible well because bg-color were set on 4 for dark and light style. Only for dark style are setted on 2, one point of distance of the background color `bg_color(3)` like light style.
This fix #822

### DIFF
--- a/src/widgets/_sidebars.scss
+++ b/src/widgets/_sidebars.scss
@@ -12,6 +12,9 @@
     &.highlight {
         background-color: bg_color(4);
         color: $fg-color;
+        @if $color-scheme == "dark" {
+            background-color: bg_color(2);
+        }
     }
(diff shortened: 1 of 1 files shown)


danrabbit:
Hey thanks for your branch! I think adding specific workarounds is something I'd like to avoid as much as possible. I have instead proposed a branch that adjusts the background level values to have a closer constrast spread with the light style which also resolves the issue

okitoxo:
Thank for you, this is an awesome project. I will continuing to try help you!

PR #970, opened 2021-03-11
Add and use unfocused selected colors

### LINKED ISSUE #734: greenfield: Should export the same public colors as Adwaita

danrabbit:
Fixes #734 

text selections now lose color in backdrop. Gtk widget factory is probably one of the easiest ways to test this

### DIFF
--- a/src/_exported.scss
+++ b/src/_exported.scss
@@ -26,6 +26,8 @@
 
 @define-color selected_bg_color #{'alpha(@accent_color, 0.25)'};
 @define-color selected_fg_color #{'shade(@accent_color, 0.5)'};
+@define-color theme_unfocused_selected_bg_color #{'alpha(@text_color, 0.10)'};
+@define-color theme_unfocused_selected_fg_color #{"" + $fg-color};
 
 @if $color-scheme == "dark" {
     @define-color menu_separator #{rgba(black, 0.25)};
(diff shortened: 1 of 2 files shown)


cassidyjames (approved):
This seems on the lower-contrast side in the dark style, but we can iterate.

PR #971, opened 2021-03-12
Index: Adjust dark levels by contrast

### LINKED ISSUE #822: Sidebar selection not visible in dark style

danrabbit:
Supersedes #965
Fixes #822

Adjusted level 3 and 4 in the dark style so that we have the same contrast ratio between steps as we do in the light style. Ever so slightly darken titlebars, but using level4 feels really harsh in the dark style

### DIFF
--- a/src/_index.scss
+++ b/src/_index.scss
@@ -17,7 +17,7 @@ $warning_color: $BANANA_900;
 $warning-icon-color: mix($BANANA_500, $BANANA_700, $weight: 50%);
 
 @if $color-scheme == "dark" {
-    $titlebar-color: mix($BLACK_500, $BLACK_700, $weight: 70%);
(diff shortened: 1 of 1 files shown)


cassidyjames (approved):
This feels a bit on the too-high-contrast side, but it's an improvement. I almost wonder if there's a perception thing with darker colors _seeming_ higher contrast or something. But if you're happy, I'm happy.

danrabbit:
Yeah I think there might be a perception thing going on for sure. Depending on the environment it's either "OMFG THE DARKNESSSS" or like so subtle. I think this is better, but I probably am relying on the math too much. So we can continue to iterate!

PR #973, opened 2021-03-12
Buttons: suggested-action extends selection

### LINKED ISSUE #866: Suggested action buttons don't follow app accent color
## What Happened

I have been playing with setting the accent color within apps for elementary OS 6. However, suggested action buttons seem to always follow the system accent color instead of the in-app defined accent color. Do apps need to define all the shades as well?

## Expected Behavior

Setting `@define-color accent_color #f00` would set the color for suggested action buttons.

## Steps to Reproduce

1. Set the `@accent_color` for an app in its CSS
2. Use a suggested action button
3. See that it still follows the system accent color

## Platform Information

elementary OS 6 daily with stylesheet from git master

danrabbit:
Fixes #866 

Added benefit of less code and matching selection styles. Even if its too pastel in the light style or we like the brighter text in the dark style, I think this is the right direction to go in code and we can adjust the selection style to be more saturated if that's what we want

## BEFORE
Screenshot from 2021-03-11 17 04 59@2x
Screenshot from 2021-03-11 17 05 05@2x

## AFTER
Screenshot from 2021-03-11 17 04 11@2x
Screenshot from 2021-03-11 17 04 25@2x

### DIFF
--- a/src/widgets/_buttons.scss
+++ b/src/widgets/_buttons.scss
@@ -57,36 +57,28 @@ button {
         }
 
         &.suggested-action:not(:disabled) {
-            $accent-fg-color: #{'mix (black, @accent_color_900, 0.8)'};
-            $bg-color: #{'@accent_color_100'};
-            $border-color: #{'alpha (mix (black, @accent_color_300, 0.8), 0.4)'};
-            $checked-border-color: alpha #{'(@accent_color_500, 0.45)'};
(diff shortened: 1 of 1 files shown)


cassidyjames (approved):
@danrabbit yeah I'd definitely like to bring up the saturation, then, but this is the right direction.

PR #975, opened 2021-03-12
Darken selection in light style

danrabbit:
Follow up for #973 

A little bit less pastel

### DIFF
--- a/src/_exported.scss
+++ b/src/_exported.scss
@@ -24,8 +24,8 @@
 @define-color menu_separator #{rgba(black, 0.15)};
 @define-color menu_separator_shadow #{rgba(white, 0.8)};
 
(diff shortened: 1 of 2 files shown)


danrabbit:
TBH, I kinda like the more pastel buttons @cassidyjames. Maybe sit on this a few days? :)

cassidyjames:
I definitely prefer the less pastel look here, especially across all the different accent colors. But I can sit on it.

hanaral:
I think my preference is the same with Cassidy's. You have to account at least a little bit for the lighter pastel-y colours getting clipped on generic low quality displays. Optimally they would all be viewed on P3 displays but that aint happening any time soon :^)

PR #983, opened 2021-04-04
Improve Notebook styling

jepotter1-archive:
With light theme:

notebook-files

With dark theme:

notebook-code

Works with notebooks of all orientations (top, bottom, left & right):

gtk-widget-factory-notebook

### DIFF
--- a/src/widgets/_notebooks.scss
+++ b/src/widgets/_notebooks.scss
@@ -3,25 +3,30 @@ notebook {
         background-color: bg-color(3);
 
         $gradient: 
-            rgba(black, 0),
-            rgba(black, 0) calc(100% - #{rem(2px)}),
-            rgba(black, 0.05) calc(100% - #{rem(1px)}),
-            rgba(black, 0.15);
+            rgba(black, 0.07) 0%,
+            rgba(black, 0.03) 20%,
+            rgba(black, 0.03) 80%,
+            rgba(black, 0.07) 100%;
+        $border: 1px solid rgba(black, 0.3);
 
         &.top {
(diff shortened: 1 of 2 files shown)


jwinnie:
Hmmm... I'm going to have to think about the design of this a bit more.

@elementary/ux design feedback/input needed

jwinnie:
Okay, I've redesigned it to make it fit more with elementary's overall style, and also to improve on their sense of "physicality":

notebook-files

notebook-code

gtk-widget-factory-notebook

hanaral:
I'm pretty sure this was an intentional change, and I feel that trying to rewrite something that was only recently done is not the right path. 
If you want to give connected tabs a shot I'd recommend on top of the sass Cassidy was experimenting with in this [branch](https://github.com/elementary/stylesheet/tree/notebook-traditional)

jwinnie:
> I'm pretty sure this was an intentional change, and I feel that trying to rewrite something that was only recently done is not the right path.
> If you want to give connected tabs a shot I'd recommend on top of the sass Cassidy was experimenting with in this [branch](https://github.com/elementary/stylesheet/tree/notebook-traditional)

I tried that branch, but unfortunately it produces a bunch of weird visual artifacts and buginess when used with latest master, so I thought it would be easier just to write it from scratch.

hanaral:
> I tried that branch, but unfortunately it produces a bunch of weird visual artifacts and buginess when used with latest master, so I thought it would be easier just to write it from scratch.

All you have to do is copy and paste the part that does the connected style on to the up to date sass.

PR #989, opened 2021-04-06
Support Hdy.Tabbox

danrabbit:
This theoretically supports Hdy.Tabbox, but there's some funky rendering issues that I don't think are my fault


danrabbit:
Opened an issue upstream here: https://gitlab.gnome.org/GNOME/libhandy/-/issues/427

danrabbit:
@cassidyjames so I opened an issue upstream, but it sounds like it's not gonna get fixed. So the only thing to do here is just to not use box shadow. Which kind of blows, but it is what it is I guess. This is at least not totally broken

Issue #992, opened 2021-04-19
Hdy.Tab indicator should be flat(ter)


cassidyjames:
## What Happened

In the Handy tab view demo, if you enable an indicator on a tab you get a raised button on top of the tab, which adds padding and borders that don't look great.

No indicator | Indicator
--- | ---
![Screenshot from 2021-04-19 13-20-21](https://user-images.githubusercontent.com/611168/115291585-2a887000-a112-11eb-96ae-8436ca447670.png) | ![Screenshot from 2021-04-19 13-20-16](https://user-images.githubusercontent.com/611168/115291583-29efd980-a112-11eb-9294-b9a54ceb2252.png)

## Expected Behavior

The indicator is a flat button, like in Adwaita.

No indicator | Indicator
--- | ---
![Screenshot from 2021-04-19 13-21-04](https://user-images.githubusercontent.com/611168/115291647-3d9b4000-a112-11eb-8b9e-e22227780189.png) | ![Screenshot from 2021-04-19 13-21-09](https://user-images.githubusercontent.com/611168/115291649-3d9b4000-a112-11eb-9c46-9e3e0092f8fa.png)

## Steps to Reproduce

1. Open Handy Demo's tab view demo
2. Right-click a tab and check the "Indicator" item

## Platform Information

elementary OS 6 Early Access with Stylesheet from master

PR #1012, opened 2021-05-04
Extend .flat class to Hdy.Tab indicator buttons

alvvaro:
smol fix to #992

Screenshot from 2021-05-04 13 17 11

### DIFF
--- a/src/widgets/_notebooks.scss
+++ b/src/widgets/_notebooks.scss
@@ -68,7 +68,8 @@ tab {
     min-width: 24px;
     padding: rem(3px);
 
-    .tab-close-button, .tab-indicator{
+    .tab-close-button,
+    .tab-indicator {
         @extend .flat;
     }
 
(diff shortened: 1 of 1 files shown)


src/widgets/_notebooks.scss
```diff
@@ -68,7 +68,7 @@ tab {
     min-width: 24px;
     padding: rem(3px);
 
-    .tab-close-button {
+    .tab-close-button, .tab-indicator{
```
cassidyjames, line 71: Thanks, this was simple. :sweat_smile: One style nit: we put selectors on their own lines, and leave spaces between selectors and the opening brace.

```suggestion
    .tab-close-button,
    .tab-indicator {
```

cassidyjames (changes requested):

src/widgets/_notebooks.scss
alvvaro: Yeah, that way looks clearer. Thanks for the tip!

danrabbit (approved):
Nice fix! Thanks for taking care of that :)

PR #1025, opened 2021-05-12
Notebooks: use tab_base_color

### LINKED ISSUE #1024: Add support for notebook.view
We used to use `.inline-toolbar` for the more seamless tab design, but since inline toolbars is being dropped, we should support this with a style class like `.view`. (See: #379 from the previous stylesheet)

danrabbit:
Fixes #1024 

No need for another style class anymore. Also self-contain hdy.tabbox a bit here to fix issues with box shadow


cassidyjames (approved):
I like

PR #1035, opened 2021-05-14
Menus: Fix decoration selector

### LINKED ISSUE #791: Menu shadows are too large
Something has gone awry and menu shadows are now as big as window shadows and there is no difference for submenu shadows

danrabbit:
Fixes #791

### DIFF
--- a/src/widgets/_menus.scss
+++ b/src/widgets/_menus.scss
@@ -96,14 +96,14 @@ window.popup {
         font: initial;
     }
 
-    decoration {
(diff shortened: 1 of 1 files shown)


cassidyjames:
Hm, I'm not able to reproduce the original issue?

danrabbit:
I'm guess this got fixed accidentally breaking up the window and HeaderBar stuff. Probably dropped a style that was overly specific

PR #1052, opened 2021-06-07
Scales: Add focus styles

### LINKED ISSUE #1051: Scales lack focus styles
Like #1049 but for scales. I imagine the same sort of glow on the handle would make sense?

danrabbit:
Fixes #1051 

It looks like outline doesn't affect switch sliders, so we have to use box-shadow here and do some funky layering to emulate it


cassidyjames (approved):
:+1: looks nice

PR #1089, opened 2021-09-18
Hdy.TabBar styling fixes

davidmhewitt:
This picks up the `.start-action` and `.end-action` classes and also ensures they have a correctly styled background. These are the containers on either end of the TabBar that optionally exist to hold static widgets (e.g. the new tab button and tab overflow dropdown in Ephy).

There's still an issue where the left/right borders of the first/last tabs aren't drawn. This can be fixed by adding a 1px left/right margin to all tabs, but I presume we don't want this for stylistic reasons. Not savvy enough with CSS to figure out a different fix.

The `.start-action` and `.end-action` stuff can be tested with https://github.com/elementary/browser/pull/41 , the tab border stuff is visible in the Handy example app.

### DIFF
--- a/src/widgets/_notebooks.scss
+++ b/src/widgets/_notebooks.scss
@@ -6,7 +6,7 @@
     }
 
 
-    $gradient: 
+    $gradient:
         rgba(black, 0),
         rgba(black, 0) calc(100% - #{rem(2px)}),
         rgba(black, 0.05) calc(100% - #{rem(1px)}),
@@ -101,10 +101,12 @@ tab {
     }
 }
 
-tabbox {
-    @extend %tabheader;
+tabbar {
+    .box, .start-action, .end-action {
+        @extend %tabheader;
+    }
 
-    tab {
+    .box tab {
         background-clip: padding-box;
         border: 1px solid transparent;
         margin: rem(3px) 0;


cassidyjames:
Putting a margin-left/right on the tabs seems to alleviate the clipping, but we should be smart about making sure it's on `:first-child` and `:last-child`. Something to watch out for is when we have pinned tabs; you can easily end up with double margins there because they're in separate boxes, so the last pinned tab and the first non-pinned tab both put margins between them. Not sure the best way to fix that.

I also think we can/should be using the `box` name instead of `.box` class, though I'm not sure it matters in practice (both seem to work the same here).

davidmhewitt:
I tried the `:first-child` / `:last-child` thing but noticed it didn't seem to update properly when pinning and unpinning tabs, so you'd still end up with the double spacing thing.

davidmhewitt:
Oh, but those things are not necessarily related. I'm being dumb.

The problem that I was seeing that adding a left margin to the first child worked fine until you pinned and unpinned a tab. Then the margin didn't get re-applied to the newly unpinned tab, causing it to have clipping again.

cassidyjames:
Ah yeah I can reproduce the weirdness with pinning/unpinning and also just reordering. It's like reordering doesn't trigger a style update so the tabs don't get the new `:first-child` styles.

As a workaround for now, I'd be fine adding a 1px margin on the left and right and including a comment as to why we're doing that.

davidmhewitt:
@cassidyjames Done. I've also kicked off a rebuild of https://github.com/elementary/browser/pull/41 with this branch included, so you can download the bundle there in 15 mins or so and test out these changes if you wish.

cassidyjames:
Alright I think the last two things that are really sticking out to me:

1. We lost the shadow at the bottom of the headerbar compared to befoe
2. Pinned tabs are shorter than non-pinned tabs

Screenshot from 2021-09-22 19 54 01 | Screenshot from 2021-09-22 19 54 10
---|---

davidmhewitt:
I think I'm seeing that `Hdy.TabBar` doesn't have/support the `.bottom`, `.left`, `.right` classes, which is probably what's causing the shadow issue.

The tab height thing exists in the current stable version of epiphany too.

Either way, I really don't know enough about the best way to do things in CSS to even attempt fixing those :sweat: 

I have this branch set to allow pushes from contributors, so, please feel free to hack on this.

Issue #1091, opened 2021-10-06
Gtk.Separator nearly invisible in dark mode in menus and sidebars


jeremypw:
### What Happened?

The separator line used in context menus and (e.g.) Files sidebar is hardly visible when in dark mode.  See e.g. https://github.com/elementary/files/issues/1856

See also example screenshots below.

### Steps to Reproduce

Files Sidebar
![DarkModeSeparatorSidebar](https://user-images.githubusercontent.com/10513844/136242469-19b8de8a-6ebd-40c8-a405-5e37e205e859.png)

Files Context Menu
![DarkModeContextMenuSeparator](https://user-images.githubusercontent.com/10513844/136242486-84581469-82f4-4c2e-975a-7b61cd06560f.png)

Appearance of context menu separator similar in e.g. Terminal

### Expected Behavior

The separator should be equally visible in light and dark mode.  As the text in dark mode is paler than the background in dark mode it would be expected that the separators would also be paler (which would improve visibility)

### OS Version

6.x (Odin)

### Software Version

Compiled from git

### Log Output

```shell
N/A
```

### Hardware Info

N/A

PR #1093, opened 2021-10-14
adjust opacity of menu_separator in dark color scheme

a01wielo:
This adjustment makes the horizontal separation line a bit darker, so it can be distinguished more clearly from the background color of the dark theme. See issue #1091 and issue elementary/files#1856


danrabbit (approved):
Subtle change but looks good to me!

Here's a direct comparison:
Screenshot from 2021-11-23 11 03 56
Screenshot from 2021-11-23 11 03 22

PR #1101, opened 2021-11-16
Use the accent color for highlighting in IBus candidate window

### LINKED ISSUE #1067: IBus-pinyin popover shows cursor as black rectangle, obscuring the text
## What Happened

 - Installed the "Intelligent Pinyin" input method.
 - Selected "Chinese (Intelligent Pinyin) using Wingpanel
 - Opened Code and typed (e.g.) "xi" into new document
 - The pinyin popover opens - the first option is obscured.

IntelligentPinyin

Note that carrying out the same steps on Ubuntu Focal with the default theme gives the following:

Screenshot from 2021-08-02 18 36 38

This indicates the issue is (probably) caused by the elementary theming.

## Expected Behavior

The cursor is displayed without obscuring the text.

## Platform Information

Odin daily.

ryonakano:
Fixes #1067

## Before
2021-11-16 12 07 05 のスクリーンショット

The highlight word in IBus candidate window is redacted and unreadable.

## After
2021-11-16 12 05 34 のスクリーンショット

Now the highlight word uses the accent color preferred in System Settings, so it's visible and plus provides design consistency.

I'm not familiar with the structure of this repository, so feel free to move the changes to the appropriate place.

### DIFF
--- a/src/widgets/_views.scss
+++ b/src/widgets/_views.scss
@@ -42,6 +42,7 @@ scrolledwindow {
 .view {
     background: #{'@base_color'};
     color: $fg-color;
+    -gtk-secondary-caret-color: #{'@accent_color_100'};
 
     &:disabled {
         background: $insensitive-base-color;
--- a/src/widgets/_views.scss
+++ b/src/widgets/_views.scss
@@ -42,7 +42,7 @@ scrolledwindow {
 .view {
     background: #{'@base_color'};
     color: $fg-color;
-    -gtk-secondary-caret-color: #{'@accent_color_100'};
+    -gtk-secondary-caret-color: #{'@selected_bg_color'};
 
     &:disabled {
         background: $insensitive-base-color;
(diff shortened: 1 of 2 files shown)


jeremypw (approved):
I can confirm that this fixes the issue for both Chinese Intelligent Pinyin and Japanese-Mozc - nice work!  The fix works in both light and dark themes (the candidate windows are not dark preference aware)

src/widgets/_views.scss
```diff
@@ -42,6 +42,7 @@ scrolledwindow {
 .view {
     background: #{'@base_color'};
     color: $fg-color;
+    -gtk-secondary-caret-color: #{'@accent_color_100'};
```
danrabbit, line 45: this should probably instead be:
```suggestion
    -gtk-secondary-caret-color: #{'@selected_bg_color'};
```

danrabbit (changes requested):
Nice work figuring out the fix for this! Just one small requested change

src/widgets/_views.scss
ryonakano: Oh there is a dedicated variable already, thank you for your suggestion!

danrabbit (approved):

PR #1108, opened 2021-11-25
Minimum support for GTK4

danrabbit:
Allows us to build a super broken-ass version of our stylesheet for GTK4

Screenshot from 2021-11-24 17 35 43@2x


cassidyjames (approved):
Everything appears to still work as expected in GTK3. LFG

PR #1112, opened 2021-11-27
GTK4/Headerbars: .titlebutton → windowcontrols

danrabbit:
.titlebutton isn't a thing anymore. Instead there's an element `windowcontrols` which contains all the title buttons

There's also no longer a separator in headerbar, so to maintain margins here we can add margin to the `windowcontrols` widget

## BEFORE
Screenshot from 2021-11-26 18 06 33@2x

## AFTER
Screenshot from 2021-11-26 18 06 11@2x

### DIFF
--- a/src/gtk-4.0/widgets/_buttons.scss
+++ b/src/gtk-4.0/widgets/_buttons.scss
@@ -99,12 +99,6 @@ button {
             }
         }
 
-        &.titlebutton:not(:active) {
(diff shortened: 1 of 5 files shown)


Exalm:
Note that GtkWindowControls is a public widget. It doesn't have to be inside a header bar.

cassidyjames (approved):

PR #1113, opened 2021-11-28
GTK4/Windows,Headerbars: fix decorations for dialogs

danrabbit:
There's no longer separate element names for different types of windows. Instead, a window is a window and there are different classes (makes total sense!). This branch fixes decorations for SSD and CSD dialogs, and makes some progress (but not perfect) towards `window.dialog.message` support.

We intentionally hide titles in dialogs in the GTK3 version of our stylesheet already. Since we have a Granite.Dialog/Granite.MessageDialog I'm not sure if we want to stop doing that?

## BEFORE
Screenshot from 2021-11-28 14 45 56@2x
Screenshot from 2021-11-28 14 45 50@2x
Screenshot from 2021-11-28 14 45 45@2x
Screenshot from 2021-11-28 14 45 39@2x

## AFTER
Screenshot from 2021-11-28 14 45 04@2x
Screenshot from 2021-11-28 14 44 57@2x
Screenshot from 2021-11-28 14 44 52@2x
Screenshot from 2021-11-28 14 44 45@2x

### DIFF
--- a/src/gtk-4.0/widgets/_headerbars.scss
+++ b/src/gtk-4.0/widgets/_headerbars.scss
@@ -139,18 +139,5 @@ window {
 }
 
 window.dialog.csd > .titlebar {
-   background: inherit;
-   box-shadow: outset-highlight("top");
-   padding: rem(6px);
-
-   &.default-decoration {
-       padding: 0;
-   }
-
-   label.title {
-       text-shadow: none;
-       -gtk-icon-shadow: none;
-       font-size: 0.1px; // Workaround for an issue in Gtk when setting 0px
-       color: transparent;
-   }
+    @extend .flat;
 }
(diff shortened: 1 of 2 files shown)


cassidyjames:
>We intentionally hide titles in dialogs in the GTK3 version of our stylesheet already. Since we have a Granite.Dialog/Granite.MessageDialog I'm not sure if we want to stop doing that?

Yeah I'd be fine with that; if you don't want a titlebar, use a dialog that doesn't have a titlebar (or remove it yourself). :) Probably better for cross-platform expectations anyway, as apps not designed for elementary OS should look more like how their developers expect.

danrabbit:
@cassidyjames alright done. Now we just treat dialog titlebars as flat titlebars

cassidyjames (approved):

PR #1116, opened 2021-11-29
GTK4/Typography: support gtk title classes

### LINKED ISSUE #518: Add new Gtk text style classes

danrabbit:
Fixes #518 
Fix an issue with dialog titles in headerbars

These styles are used in GTK4 Widget Factory, but afaict the only one used in GTK itself is `.dialog .title`.  There's no harm in matching API with Adwaita here though


BrainBlasted:
>  The way we're using .h1 seems to be the same as .large-title. I'm not sure if this means we should make those the same thing or bump things up so that .h2 is now .title-1 etc

Something to consider - `.large-title` will probably be deprecated upstream, as no one uses it in GTK and libadwaita now uses `title-1`.

danrabbit:
@BrainBlasted thats good to know! Thanks!

cassidyjames (approved):

PR #1124, opened 2022-01-03
GTK4: Replace icon-effect with filter

danrabbit:
This is based on how Adwaita replaced `-gtk-icon-effect` which no longer exists

See for example: https://github.com/GNOME/libadwaita/search?l=SCSS&q=icon+dim


cassidyjames (commented):
Looks like we're using `-` separation in other variable names, should we do the same here?

danrabbit:
@cassidyjames oops you're right. Fixed!

cassidyjames (approved):

PR #1152, opened 2022-01-05
GTK4: style ShortcutWindow

danrabbit:
Screenshot from 2022-01-05 09 59 56

There's still not toplevel selector, so we can't do anything about the headerbar style :(

* Fix window margin on Gtk.ShortcutWindow
* Remove `+` signs from Gtk.ShortcutLabel

### DIFF
--- a/src/gtk-4.0/widgets/_shortcuts.scss
+++ b/src/gtk-4.0/widgets/_shortcuts.scss
@@ -6,11 +6,6 @@ shortcuts-group > label {
     @extend .h4;
 }
 
-// Remove plus signs in Gtk.ShortcutLabel
-shortcut .dim-label {
-    color: transparent;
-}
-
 accelerator,
 .keycap {
     background-color: bg-color(2);
--- a/src/gtk-4.0/widgets/_shortcuts.scss
+++ b/src/gtk-4.0/widgets/_shortcuts.scss
@@ -6,6 +6,12 @@ shortcuts-group > label {
     @extend .h4;
 }
 
+// Characters between keys in Gtk.ShortcutLabel, like + or /
+ shortcut .dim-label {
+     margin-left: rem(3px);
+     margin-right: rem(3px);
+ }
+ 
 accelerator,
 .keycap {
     background-color: bg-color(2);
--- a/src/gtk-4.0/widgets/_shortcuts.scss
+++ b/src/gtk-4.0/widgets/_shortcuts.scss
@@ -1,0 +1,42 @@
+shortcuts-section {
+    margin: rem(12px);
+}
+
+shortcuts-group > label {
+    @extend .h4;
+}
+
+// Remove plus signs in Gtk.ShortcutLabel
+shortcut .dim-label {
+    color: transparent;
+}
+
+accelerator,
+.keycap {
+    background-color: bg-color(2);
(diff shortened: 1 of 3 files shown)


src/gtk-4.0/widgets/_shortcuts.scss
```diff
@@ -0,0 +1,42 @@
...
+    margin: rem(12px);
+}
+
+shortcuts-group > label {
+    @extend .h4;
+}
+
+// Remove plus signs in Gtk.ShortcutLabel
+shortcut .dim-label {
+    color: transparent;
```
cassidyjames, line 11: Do we also want to change the font size here so the spacing between keycaps iisn't determined by the invisible `+`?

danrabbit: We could! One character width seems to be a good spacing already to me, but I'm open to suggestions

cassidyjames: Okay, maybe this is a hack and not worth it, but then again hiding the `+` in the first place feels like a hack, so… :shrug: We use 6px column spacing in our shortcut overlay, for example:

```suggestion
    color: transparent;
    font-size: 0;
    margin-left: rem(3px);
    margin-right: rem(3px);
```

danrabbit: Would it make more sense to just make the font size rem(6px)?

cassidyjames (changes requested):
On second thought, hiding dim labels in CSS doesn't seem like something we want to do; it looks like some apps also use `/` in dim labels to represent multiple shortcuts, e.g.

><kbd>Super</kbd>+<kbd>Space</kbd> / <kbd>Alt</kbd>+<kbd>F2</kbd>

With this styling, it would not be clear that was two different shortcuts. So, I guess we just leave the dim labels alone in GTK shortcut windows?

src/gtk-4.0/widgets/_shortcuts.scss
cassidyjames: @danrabbit probably not because the `+` character isn't guaranteed to actually be `1em` in size, so it wouldn't match anyway. But see my requested changes; I don't think we want to actually fiddle with the styling of dim labels, anyway, due to how they're actually used.

src/gtk-4.0/widgets/_shortcuts.scss
```diff
@@ -0,0 +1,37 @@
+shortcuts-section {
+    margin: rem(12px);
+}
+
+shortcuts-group > label {
+    @extend .h4;
+}
+
```
cassidyjames, line 8: Okay sorry for this back and forth but it looked wrong with the labels butting up against the keys. I think this looks pretty correct?

```suggestion

// Characters between keys, like + or /
 shortcut .dim-label {
     margin-left: rem(3px);
     margin-right: rem(3px);
 }
 
```

Screenshot from 2022-01-11 15 12 54

cassidyjames (changes requested):
Okay one tiny thing after looking at it locally, but I think over-all this is probably a good start.

danrabbit:
No worries, good change 👍

cassidyjames (approved):

PR #1161, opened 2022-01-12
GTK4/Popovers: fix modelmenu styles

danrabbit:
This fixes all the modelbutton styles in GTK4 Widget Factory, but only for LTR. It looks like RTL has a lot of other issues that I'd like to follow up on in future PRs

## BEFORE

Screenshot from 2022-01-12 11 43 38
Screenshot from 2022-01-12 11 43 28
Screenshot from 2022-01-12 11 43 13
Screenshot from 2022-01-12 11 43 00

## AFTER
Screenshot from 2022-01-12 11 42 26
Screenshot from 2022-01-12 11 42 03
Screenshot from 2022-01-12 11 41 46
Screenshot from 2022-01-12 11 41 30

### DIFF
--- a/src/gtk-4.0/widgets/_popovers.scss
+++ b/src/gtk-4.0/widgets/_popovers.scss
@@ -199,6 +199,43 @@ popover.background {
         separator {
             margin: rem(3px) 0;
         }
+
+        // ModelMenu
+        box.circular-buttons,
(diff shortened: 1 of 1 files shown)


cassidyjames (approved):
This looks so much nicer! Nice work!

PR #1164, opened 2022-01-12
GTK4: fix link styles

danrabbit:
* Make sure we don't style the entire label with accent color when it contains a link
* Add underline to links

### DIFF
--- a/src/gtk-4.0/_typography.scss
+++ b/src/gtk-4.0/_typography.scss
@@ -39,9 +39,7 @@ label.h4,
     font-size: 0.85em;
 }
 
-.accent,
-button.link,
-link {
+.accent {
     @if $color-scheme == "light" {
         color: #{'shade(@accent_color, 0.75)'};
     } @else if $color-scheme == "dark" {
@@ -136,6 +134,7 @@ link {
 
 button.link,
 link {
+    @extend .accent;
     text-decoration: underline;
 }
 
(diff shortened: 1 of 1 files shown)


src/gtk-4.0/_typography.scss
```diff
@@ -40,8 +40,8 @@ label.h4,
 }
 
 .accent,
```
cassidyjames, line 42:
```suggestion
.accent {
```

src/gtk-4.0/_typography.scss
```diff
@@ -40,8 +40,8 @@ label.h4,
...
-.link,
-:link {
+button.link,
+link {
```
cassidyjames, line 44:
```suggestion
```

src/gtk-4.0/_typography.scss
```diff
@@ -134,6 +134,11 @@ label.h4,
     }
 }
 
+button.link,
+link {
+    text-decoration: underline;
```
cassidyjames, line 139:
```suggestion
    @extend .accent;
    text-decoration: underline;
```

cassidyjames (changes requested):
So this is a tiny thing, but keeps the Sass more in sync in my experience: you can use `@extend` to ensure one selector stays in sync with the other, which means you don't have to make sure to update/change multiple places in the future.

It's optional, but in reading this change it stood out to me.

danrabbit:
@cassidyjames I've had some issues with extends pulling in styles where I don't want them, but this seems reasonable :+1:

cassidyjames (approved):

Issue #947, opened 2021-02-02
Accent colour in the volume slider


Rantyrant:
## Problem

In the sound indicator, the volume slider uses a greyish colour as shown in this image on elementary 5:
![image](https://user-images.githubusercontent.com/65864655/106592598-03c58f00-6550-11eb-9c8e-43909b70ff39.png)
It is distinguishable enough on the light theme in elementary 5, but in the early access of elementary 6.0 that volume slider kind of blends into the dark theme, it probably depends on the quality of monitor, but on my old laptop where I tested the early access, it was hard to make out (didn't manage to get a screenshot of this before wiping)

## Proposal

I propose the volume slider uses the accent colours you select yourself during the first time run of elementary 6.0.

## Prior Art

I suck at making mockups but it would be something akin to what's done i Cinnamon:

![cinnamon](https://user-images.githubusercontent.com/65864655/106593461-2015fb80-6551-11eb-82ae-536ed9042434.png)

danrabbit:
I’m gonna transfer this to the stylesheet so we can look at doing this for all scales and not just here in the panel. I believe the `accent` class already works for scales, but if it’s not too visually disruptive it could bring a nice pop of color to always use accent color for them

danrabbit:
Couple of examples of scales using accent color

![Screenshot from 2021-08-17 12 51 01](https://user-images.githubusercontent.com/7277719/129791177-2e38e7ab-26cd-4c21-bc99-d8278db0cb94.png)
![Screenshot from 2021-08-17 12 50 24](https://user-images.githubusercontent.com/7277719/129791180-5a62fd0b-17bf-48fb-a785-caa02d447bf2.png)

cassidyjames:
Hm, not super opposed. If people want an accent-color-free experience, they can choose Slate. ;)

rakshit087:
Is this issue resolved? I am looking for some beginner friendly issues to contribute to.

PR #1211, opened 2022-06-14
use accent_color in highligth for all scales

KaltWulx:
Allows the user to change the highlight color for all scales with "accent_color", initially intended to resolve issue #947 

Result:
image


BAProductions:
Your code broken it's not #{'accent_color'}; it's #{'@accent_color'};

PR #1215, opened 2022-06-28
Index: make checkerboard much more subtle

danrabbit:
Makes the checkerboard effect much more subtle so that it interferes with image previews less


jeremypw (approved):
This improves matters significantly, thanks.

PR #1221, opened 2022-08-15
Gtk4/Scales: fix mark selector

### LINKED ISSUE #1219: GTK 4: Invisible marks on scales
### What Happened?

image

image

### Steps to Reproduce

1. Look at scale

### Expected Behavior

Marks should be shown

### OS Version

7.x (Early Access)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

danrabbit:
Fixes #1219

It looks like vertical scale marks are always on the right, even in RTL so we don't need to account for that

### DIFF
--- a/src/gtk-4.0/widgets/_scales.scss
+++ b/src/gtk-4.0/widgets/_scales.scss
@@ -22,8 +22,8 @@ scale {
         }
     }
 
-    mark indicator {
-        color: mix(bg-color(4), $fg-color, $weight: 75%);
+    mark {
+        background: mix(bg-color(4), $fg-color, $weight: 75%);
     }
 
     slider {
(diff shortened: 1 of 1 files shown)


lenemter (changes requested):
Vertical scale marks are not always on the right, because you can specify their position in add_mark method.
Marks on the left have no margin

image

danrabbit:
Ah sorry you're absolutely right. Thanks for the attention to detail!

danrabbit:
@lenemter should be fixed now

lenemter (approved):
Fixes the issue now. Thanks

danrabbit:
Thanks for reviewing!

PR #1222, opened 2022-08-28
HeaderBars: set large-icons size to 24px

danrabbit:
Makes my life easier to set icon sizes for buttons and menu buttons in titlebars. Especially for menubuttons since manually setting the child means no image-button class and it isn't easy to select the button since the popover is also a child. `has_frame` technically works here, but it adds `flat` which is slightly different from `image-button`.

Also in `images`, don't require the `large-icons` class to be directly on an image, because it's useful to set it on a container widget like a button

Vala before:
```vala
        var updates_button = new Gtk.Button.from_icon_name ("software-update-available");
        ((Gtk.Image) updates_button.child).pixel_size = 24;

        var menu_button = new Gtk.MenuButton () {
            child = new Gtk.Image.from_icon_name ("open-menu") {
                pixel_size = 24
            },
            has_frame = false
        };
```

Vala after:
```vala
        var updates_button = new Gtk.Button.from_icon_name ("software-update-available");
        updates_button.add_css_class ("large-icons");

        var menu_button = new Gtk.MenuButton () {
            icon_name = "open-menu",
        };
        menu_button.add_css_class ("large-icons");
```

### DIFF
--- a/src/gtk-4.0/widgets/_headerbars.scss
+++ b/src/gtk-4.0/widgets/_headerbars.scss
@@ -45,7 +45,6 @@ window {
             -gtk-icon-shadow: none;
         }
 
-        &.large-icons,
         .large-icons {
             -gtk-icon-size: 24px;
         }
--- a/src/gtk-4.0/widgets/_headerbars.scss
+++ b/src/gtk-4.0/widgets/_headerbars.scss
@@ -44,6 +44,11 @@ window {
             text-shadow: none;
             -gtk-icon-shadow: none;
         }
+
+        &.large-icons,
+        .large-icons {
+            -gtk-icon-size: 24px;
+        }
     }
 
     &.csd {
--- a/src/gtk-4.0/widgets/_images.scss
+++ b/src/gtk-4.0/widgets/_images.scss
@@ -7,12 +7,12 @@ image {
     &:disabled {
         filter: opacity($disabled-opacity);
     }
+}
 
-    &.normal-icons {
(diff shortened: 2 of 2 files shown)


Exalm:
Isn't it easier to set menu_button.icon_size = LARGE?

Exalm:
Oh huh, my bad. That's only a thing on GtkImage, somehow I thought it was on buttons as well. Nevermind...

danrabbit:
@Exalm it looks like `icon_size` is a property for Gtk.Image but not Gtk.Button

Exalm:
Indeed. Kinda defeats the point really, I wonder if it's just an oversight.

danrabbit:
Yeah that would be pretty clean to have that as a property of Gtk.Button imo! I would :+1: that issue/PR ;)

tintou:
If you want it to be a "large-icon", I think that you need:
```
var updates_button = new Gtk.Button.from_icon_name ("software-update-available");
((Gtk.Image) updates_button.child).icon_size = Gtk.IconSize.LARGE;
```

Exalm:
The fact the button child when you use `from_icon_name()` is:

1. is not null
2. is a GtkImage

are implementation details. Don't rely on that, it can change at any time (and has changed for GtkMenuButton, where it went from menubutton > button > image to menubutton > button > box > image in 4.6).

danrabbit:
Yeah, I want to avoid trying to get and cast the child because it's not consistent or necessarily reliable

PR #1223, opened 2022-08-31
SpinButtons: give padding to direct image children

danrabbit:
Since SpinButton is no longer an Entry subclass it doesn't have the icon properties and you have to parent Images manually. This makes sure we give those Images padding like they would get in Gtk.Entry

### DIFF
--- a/src/gtk-4.0/widgets/_spinbuttons.scss
+++ b/src/gtk-4.0/widgets/_spinbuttons.scss
@@ -32,6 +32,10 @@ spinbutton {
             }
         }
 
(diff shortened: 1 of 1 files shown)


Exalm:
> and you have to parent Images manually. 

That's not something that's supposed to work really. Open an issue to have icon support in GTK?

danrabbit:
@Exalm tada: https://gitlab.gnome.org/GNOME/gtk/-/issues/5153

Exalm:
Thx, subscribed.

PR #1233, opened 2022-10-06
Fix heading label padding in lists

### LINKED ISSUE #1232: No padding on some headers
### What Happened?

image
image
image

### Steps to Reproduce

1. ???

### Expected Behavior

Padding as before

### OS Version

7.x (Early Access)

### Software Version

Latest release (I have run all updates)

### Log Output

_No response_

### Hardware Info

_No response_

danirabbit:
Fixes #1232

### DIFF
--- a/src/gtk-4.0/widgets/_lists.scss
+++ b/src/gtk-4.0/widgets/_lists.scss
@@ -17,6 +17,7 @@ listview {
         }
     }
 
+    .heading,
     .title-4,
     .h4 {
         padding-left: rem(6px);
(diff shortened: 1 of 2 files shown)


lenemter (approved):
Fixes the issue. Thanks

lenemter:
Wait I didn't notice that header in switch button at the bottom:
image

danirabbit:
@lenemter I think we should consider this a bug in the panel because it should add the `title-4` style class instead of using a headerlabel since it's not actually a header. That seems to fix it here:

Screenshot from 2022-10-06 15 57 05

lenemter:
Makes sense. You can merge this PR then

PR #1242, opened 2023-02-22
Add 3px of padding check, radio to match setting > desktop > appearance

BAProductions:
### **Before:**
Screenshot from 2023-02-21 20-30-55
### **After:**
Screenshot from 2023-02-21 20-28-22

I  also believe radio button should use a circle when checked instead of a checkmark

### DIFF
--- a/src/gtk-3.0/widgets/_checkbuttons.scss
+++ b/src/gtk-3.0/widgets/_checkbuttons.scss
@@ -1,6 +1,7 @@
 check,
 radio {
     @extend %outset-background;
+    padding: rem(3px);
     background-color: bg-color(0);
     border: 1px solid $border-color;
     box-shadow:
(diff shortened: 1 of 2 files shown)


BAProductions:
@danirabbit

danirabbit:
Hey thanks for your branch. I don't think we want to make all radio buttons this large. For specifically those color selection buttons larger radios makes sense, but for most uses cases a more compact widget is more appropriate

Closing as "Design Conflict"

BAProductions:
> Hey thanks for your branch. I don't think we want to make all radio buttons this large. For specifically those color selection buttons larger radios makes sense, but for most uses cases a more compact widget is more appropriate
> 
> Closing as "Design Conflict"

but the check make  has no padding making hard to read on 50 inch tv

PR #1245, opened 2023-03-07
Adw.PreferencesWindow

danirabbit:
Adds styles for Adw.PreferencesWindow as used in Epiphany. It isn't totally perfect and doesn't cover all the cases of Adwaita, but it makes Epiphany usable at least


src/gtk-4.0/widgets/Adwaita/List.scss
```diff
@@ -5,24 +5,47 @@ list.boxed-list {
...
         }
     }
 
     > row:not(.separator) + row:not(.separator) {
         border-top: 1px solid #{'@menu_separator'};
     }
 
+    row.entry {
+
+    }
```
Exalm, line 47: These 3 lines look like a leftover.

Exalm:
> Lists: support .boxed-list from Adw

BTW .boxed-list is in GTK now and inspector is using it, so it's not libadwaita-specific anymore. The acton rows etc still are tho.

src/gtk-4.0/widgets/Adwaita/List.scss
Exalm: Ah, no problem - just got a email notification

danirabbit:
@Exalm if you wanted to look at this and point out any obvious mistakes that would be appreciated :) It's only meant to make Adw.PreferencesWindow usable, so it doesn't address anything outside of that and I'm aware that there's some missing stuff like focus styles, error states, etc in the entry row

src/gtk-4.0/_typography.scss
```diff
@@ -155,3 +155,8 @@ label:disabled {
 .monospace {
     font-family: monospace;
 }
+
+.subtitle {
```
Exalm, line 159: Careful with this one, it is different between header bars and rows, same as with .title.

src/gtk-4.0/widgets/adw/PreferencesWindow.scss
```diff
@@ -0,0 +1,50 @@
...
+            padding-left: rem(12px);
+            padding-right: rem(12px);
+        }
+    }
+
+    preferencesgroup {
+        margin-bottom: 2rem;
+    }
+
+    leaflet stack + actionbar {
```
Exalm, line 37: Pref window doesn't have an actionbar, does it? Do you mean view switcher bar, or action bars everywhere?

src/gtk-4.0/_typography.scss
danirabbit: Good call!

src/gtk-4.0/widgets/adw/PreferencesWindow.scss
danirabbit: This targets the actionbar that appears on the passwords page

Exalm: 1. That's an ephy-specific thing. Why style it here and not in ephy?
2. That's just a regular actionbar, what does it have to do with stack/leaflet/prefwindow?

danirabbit: That's a good point. I should probably add a style sheet to our Ephy build for those kinds of specific tweaks

It was mainly just to try to make sure I'm selecting that specific actionbar instead of the view switcher one

Exalm: You know we have one upstream too, right? ;)

Exalm: Also that action bar is also there in the history window.

PR #1246, opened 2023-03-12
GTK4: Support `.keyboard-activating` class for buttons

lenemter:
See https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/3390

I guess this class needs to be used in some other places, but this is bare minimum for https://github.com/elementary/calculator/pull/242

### DIFF
--- a/src/gtk-4.0/widgets/_buttons.scss
+++ b/src/gtk-4.0/widgets/_buttons.scss
@@ -56,6 +56,7 @@ button {
                     background-clip: border-box;
                     border: 1px solid $border-color;
                 } @else if $color-scheme == "dark" {
+                    &.keyboard-activating,
                     &:active,
                     &:checked {
                         border-color: rgba(black, 0.7);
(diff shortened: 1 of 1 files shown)


BAProductions:
We need to see if other Ui Elements use .keyboard-activating  as well  & add it to the theme

BAProductions:
> See https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/3390
> 
> I guess this class needs to be used in some other places, but this is bare minimum for [elementary/calculator#242](https://github.com/elementary/calculator/pull/242)

We need to see if other UI Elements use .keyboard-activating as well & add it to the theme as needed.

Exalm:
No, it's a GtkButton thing.

lenemter:
Seems like more work needed to make this work with different styles. Converting to draft until issues resolved.

lenemter:
> Seems like more work needed to make this work with different styles. Converting to draft until issues resolved.

The issue is in https://github.com/elementary/calculator/pull/242.

Ready for review.

Exalm:
No, it doesn't. GTK3 just uses the :active state for that.

PR #1254, opened 2023-04-12
Windows: Remove Gtk3-isms and fix Mutter 44

### LINKED ISSUE #1253: Window margins cause titlebar issues in Mutter 44
In Mutter 44, X11 SSD titlebars run as a separate GTK 4 process. With our stylesheet, they look like this:

image

This is due to this margin in the stylesheet:

https://github.com/elementary/stylesheet/blob/0c841189a3ae0b06b7f5cf551d3d2c6266eda80e/src/gtk-4.0/widgets/_windows.scss#L8

If we remove this margin, it looks like this:

image

Can we have a bit of an investigation into why we have a margin here when Adwaita doesn't, and whether it's needed with a view to removing it in the next release?

danirabbit:
Fixes #1253

### DIFF
--- a/src/gtk-4.0/widgets/_windows.scss
+++ b/src/gtk-4.0/widgets/_windows.scss
@@ -5,8 +5,6 @@
 }
 
 window {
-    // Stolen from Adwaita, "lamefun trick to get rounded borders regardless of CSD use"
-    border-width: 0;
 
     &.csd {
         border-radius: rem(6px);
@@ -64,6 +62,7 @@ window {
     }
 
     &.ssd {
+        border-radius: 6px 6px 0 0;
         box-shadow: 0 0 0 1px $toplevel-border-color;
     }
 }
--- a/src/gtk-4.0/widgets/_windows.scss
+++ b/src/gtk-4.0/widgets/_windows.scss
@@ -4,71 +4,67 @@
     color: $fg-color;
 }
 
-window.csd {
-    margin: rem(12px);
-    border-radius: rem(6px);
-    box-shadow:
-        // Intentionally not in ems since it's used as a stroke
-        0 0 0 1px $toplevel-border-color,
-        // Force shadows to be the same size to prevent jumpy resize transition
-        0 13px 16px 4px transparent,
-        shadow(4);
+window {
+    // Stolen from Adwaita, "lamefun trick to get rounded borders regardless of CSD use"
+    border-width: 0;
 
-    &:backdrop {
+    &.csd {
+        border-radius: rem(6px);
         box-shadow:
+            // Intentionally not in ems since it's used as a stroke
             0 0 0 1px $toplevel-border-color,
+            // Force shadows to be the same size to prevent jumpy resize transition
             0 13px 16px 4px transparent,
-            shadow(2);
-    }
+            shadow(4);
 
-    &.maximized,
-    &.fullscreen,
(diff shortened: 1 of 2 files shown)


src/gtk-4.0/widgets/_windows.scss
```diff
@@ -4,71 +4,67 @@
...
-        margin: 0;
-    }
-
-    // black behind corners
-    &.maximized {
-        box-shadow: 0 0 0 rem(3px) black;
-    }
+window {
+    // Stolen from Adwaita, "lamefun trick to get rounded borders regardless of CSD use"
+    border-width: 0;
```
Exalm, line 9: I suspect this is a GTK3-ism too. I never figured out what it is or does and just left it in there 😅

danirabbit: Haha yeah I wasn't sure since David's screenshots from before look really odd with like spacing around the headerbar

davidmhewitt:
Mutter 44 screenshot with this branch:

image

danirabbit:
@davidmhewitt I'm sure I'll have to do something more invasive to fix those headerbar box shadows etc, but does the latest commit fix the border at least?

Exalm:
Not sure what's up with the padding around header bar...

One thing is I don't know if adwaita currently works well for Mutter - it's not using libadwaita, and I don't think it has any special case for ssd border radius either, so... yay I guess?

davidmhewitt:
Nope :confused: 

image

Adwaita comparison:

image

davidmhewitt:
> Well, that’s not Adwaita. That’s Default. Edit: i.e. what I mean is that chances are it's broken in Adwaita too and was copied here. Not sure tho, I never tested that combination (and it was impossible to test until way after that style was last touched, anyway)

Hm, for some reason, it doesn't seem to be possible to select Adwaita as a stylesheet in the inspector when you're running it against `mutter-x11-frames`.

Tbh, I don't even know what the difference between `Default` and `Adwaita` are.

Exalm:
Adwaita is a part of libadwaita. The difference:

Screenshot from 2023-04-14 01-46-18
Screenshot from 2023-04-14 01-46-35

Exalm:
Any reason to style .titlebar at all? I mean if it's not a header bar, it shouldn't be styled like a header bar then. And if it is, it's already a header bar and there won't be anything nested in there.

danirabbit:
@Exalm there are some apps that use a `box.titlebar` with a `windowcontrols` child because they're inside a `paned` instead of using a `headerbar`

Exalm:
Mm, I mean if you want to support that - ok I guess.

danirabbit:
Yeah it looks like they're just gonna have to eat shit and do something else since there's no other way to get around the padding issue for apps that nest a `headerbar` inside a `.titlebar`

Exalm:
In libadwaita we have a variant of header bar where start/end buttons can be hidden independently instead, and in the new widgets that header bar manages buttons automatically.

You could do something similar, so that dual paned apps can just use header bars instead of boxes :)

davidmhewitt (approved):
I can confirm that this resolves the issues on Mutter 44 and the issue I was seeing with Torrential. And I'm not noticing any regressions anywhere after using this branch for a bit.

But take my approval with a pinch of salt as I'm not the best at CSS or spotting minor things :sweat_smile:

PR #1257, opened 2023-04-24
Support Adw.MessageDialog

danirabbit:
Also make sure .raised overrides .flat and support .destructive

Screenshot from 2023-04-24 13 19 01

### DIFF
--- a/src/gtk-4.0/widgets/_buttons.scss
+++ b/src/gtk-4.0/widgets/_buttons.scss
@@ -34,6 +34,7 @@ button {
     &.circular.image-button,
     &.model.image-button,
     &.raised.image-button,
+    &.raised,
     & {
         @extend %outset-background;
 
@@ -178,6 +179,10 @@ button {
 
     &.text-button.link {
         padding: 0;
+    }
+
+    &.destructive {
+        @extend .destructive-action;
     }
 
     &.color {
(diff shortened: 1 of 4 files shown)


src/gtk-4.0/widgets/adw/MessageDialog.scss
```diff
@@ -0,0 +1,25 @@
...
+
+    .response-area {
+        margin-top: rem(18px);
+
+        button {
+            @extend .raised;
+        }
+
+        separator {
+            min-width: rem(6px);
```
Exalm, line 22: This looks like it will fail with vertical layout. `AdwMessageDialog` is not always horizontal, make the labels longer in inspector.

src/gtk-4.0/widgets/_buttons.scss
```diff
@@ -180,6 +181,10 @@ button {
         padding: 0;
     }
 
+    &.destructive {
```
Exalm, line 184: There's also .suggested

src/gtk-4.0/widgets/adw/MessageDialog.scss
danirabbit: Thanks for the tip! fixed :)

PR #1265, opened 2023-08-14
GTK3 menuitem: remove padding for direct children of type `label`

lenemter:
Fixes padding of items in LibreOffice. Is this an appropriate solution? This change seems to be very small but I'm not sure if it can break some apps in any way.

Before:

image

After:

image


zeebok:
I think this is more something for @elementary/ux to approve?

PR #1278, opened 2024-01-17
Buttons: Rewrite

danirabbit:
Try to rewrite with fewer overrides and selectors that are easier to follow

### DIFF
--- a/src/gtk-4.0/widgets/_buttons.scss
+++ b/src/gtk-4.0/widgets/_buttons.scss
@@ -1,41 +1,7 @@
 button {
-    background-clip: padding-box;
-    border-radius: rem(3px);
-    // Explicitly set in case containers override default color
-    color: #{'@fg_color'};
-    font-weight: 500;
-    // Off-by-one to account for padding-box clip
-    padding: rem(4px) rem(7px);
     transition: background duration("in-place") $easing;
 
-    &.flat,
-    &.image-button:not(.text-button) {
-        background: none;
-        border: none;
-        box-shadow: none;
-
-        &:checked {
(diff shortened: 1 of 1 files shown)


wpkelso:
Don't know if there is any more planned work on this, but I'm not really a fan of losing the background in the pop-over buttons. In my opinion, if we remove the background, we lose the implication that some immediate action of consequence will happen if the icon is pushed, and it parses more as a label of some sort.

wpkelso:
> Yeah this is kind of dead code. In the Granite stylesheet branch I added backgrounds for popover buttons: [elementary/granite#710](https://github.com/elementary/granite/pull/710)

Ah, is stylesheet totally integrated into granite now?  If that's the case, should this repo be archived for clarity?

---

Dunno where discussion about what convention to adopt should take place, but I suspect it's not this issue 😅

danirabbit:
> Ah, is stylesheet totally integrated into granite now?

@wpkelso no not yet! It's a goal of mine for Platform 9 but I haven't had time to work on it

wpkelso:
Oh neat! I'd love to try and help where I can then with that.

Getting back on topic, this can probably be closed then?

danirabbit:
@wpkelso I left it open just as draft in case there was any interesting code here that I might want to use later :) 

Yeah! Here's the main branch:  https://github.com/elementary/granite/pull/703

I'll try to make some time soon to write up like the main motivations and big ideas

wpkelso:
@danirabbit What are your thoughts on putting pull requests like this one under a `For Parts` or similar tag? The idea would be to easily sort out requests that are considered "dead," but might still be a source of interesting code or ideas.

PR #1279, opened 2024-01-24
Remove Settings widgets

danirabbit:
Goes with https://github.com/elementary/switchboard/pull/282


danirabbit:
I actually think I might wanna hold on this. I think I'll change the widget css name in Settings to `SettingsPage` so that it doesn't clash with `SimpleSettingsPage` and then we can remove these styles once all the panes are migrated over

danirabbit:
All the panes have been migrated and nothing should be using these styles in GTK 4 anyways

PR #1290, opened 2024-04-30
Release 8.0.0

danirabbit:


davidmhewitt:
Do we want to ship this for OS 7? Because the release branch is still set to Horus.

It also looks like 7.3 never built/released properly on launchpad too. So not sure if that needs troubleshooting first.

danirabbit:
@davidmhewitt thanks for noticing that. Updated the release branch and triggered a build for Jammy for 7.3

davidmhewitt (approved):

PR #1303, opened 2024-10-31
Popovers: use border-spacing

### LINKED ISSUE #1170: Icons in menus in GParted are cut off
### What Happened?

With GParted, the icons in the menus are cut off on the left, as shown in the following 16 second video:

https://user-images.githubusercontent.com/2998569/149635973-1ada00bd-683e-4fcb-b795-5a58074c17d3.mp4

This issue happens both in the elementary OS installer (where GParted is used for custom partitioning during install) and also if the user chooses to install GParted after install with `sudo apt install gparted`.

Whilst this may require application specific tweaks, I think that because GParted is so useful and because it is included in the installer, then it should be supported by the elementary OS theme.

### Steps to Reproduce

1. Open GParted
2. Click on the menus (e.g. the GParted or the Help menu) from the menubar.
3. The icons are cut off on the left-hand side due to their placement probably due to the stylesheet.

### Expected Behavior

The icons should not be cut off because they should be placed correctly (as they are on Gnome on Ubuntu, for example).

### OS Version

6.x (Odin)

### Software Version

Latest release (I have run all updates)

### Log Output

_No response_

### Hardware Info

The same things happens on multiple computers, so I do not think this is specific to the display or graphics hardware.

danirabbit:
Should be the actual fix for the issue in https://github.com/elementary/granite/pull/736
Fixes #1170

### DIFF
--- a/src/gtk-4.0/widgets/_popovers.scss
+++ b/src/gtk-4.0/widgets/_popovers.scss
@@ -3,6 +3,7 @@
     background: transparent;
     border: none;
     border-radius: 0;
+    border-spacing: rem(6px);
     box-shadow: none;
     color: $fg-color;
     font: initial;
@@ -48,13 +49,7 @@
 
     check,
     radio {
-        &.left {
-            margin-right: rem(6px);
-        }
-
-        &.right {
-            margin-left: rem(6px);
-        }
+        margin: 0;
     }
 
     switch slider {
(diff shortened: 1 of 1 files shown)


alainm23 (approved):

alainm23:
Good job, every day is a new learning opportunity.

PR #1319, opened 2025-04-29
Danirabbit/adw viewswitcher fixcrash

danirabbit:


ryonakano (commented):
I can confirm this does fix the crash and critical message when opening Preferences window in Web 48.0 (revision `89cd37ae96dcd9ceb35a1da1c93fbb3491373975` in `ryonakano/research-crash` branch of https://github.com/elementary/browser for anyone who want to test)

image

I'm not familiar with stylesheets at all, so needs a review from @elementary/ux in other view points e.g. if there are no regressions

wpkelso (approved):
I haven't been able to find any regressions 👍🏻

PR #1330, opened 2025-07-15
focus → focus-visible

### LINKED ISSUE #1055: Fade out focus rings after a timeout
## Problem

There’s been a number of complaints over the years from folks who don’t like seeing the keyboard focus styles all of the time. It’s probably reasonable that these styles are only important while navigating and not so important and maybe even distracting when not navigating 

## Proposal

I wonder if it might make sense to have focus styles be on a long animation so that maybe after a few seconds they slowly fade out

### LINKED ISSUE #1069: Preferences tabs focus partially stays on 'General' no matter what
## What Happened

Screenshot from 2021-08-05 22-35-22

Even when the 'Privacy' tab is in focus, the 'General' tab still gets the accent colour tint

## Expected Behavior

Both the accent colour tint and the focus shifts to other tabs when appropriate

## Steps to Reproduce

1. Open Preferences
2. Switch to another tab
3. Observe the tabs

## Platform Information

Gnome Web 3.38.3

danirabbit:
Fixes #1055
Fixes #1069

Changes use of `focus` to `focus-visible.

Widgets no longer get focus styles when clicked or on app startup, just when keyboard navigation is being used. And focus styles disappear after a little bit

### DIFF
--- a/src/gtk-4.0/widgets/_badges.scss
+++ b/src/gtk-4.0/widgets/_badges.scss
@@ -45,7 +45,7 @@
         min-height: 0;
         min-width: rem(16px);
 
-        row:selected:not(:focus) & {
+        row:selected:not(:focus-visible) & {
             color: inherit;
             background: rgba($fg-color, 0.15);
         }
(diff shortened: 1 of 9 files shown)


wpkelso (approved):
Looks alright to me 🚀

